PR #36, opened 2024-03-20
Extraction pipelines & data fetching blocks

GriffinBabe:
Implemented extraction pipelines for Sentinel2, Sentinel1 and AGERA5

The following pipelines are finished:

- [x] Sentinel-2
- [x] Sentinel-1
- [x] AGERA5


GriffinBabe:
Remaining for the Sentinel-2 Pipeline:

- [ ] Finish the memory optimization removing observations with very large high proba cloud mask coverage.
- [ ] Analyze backend behavior and optimize possible bottlenecks @jdries

GriffinBabe:
@jdries You can start profiling the backend bottlenecks within the `scripts/extraction/extract_optical.py` file in this branch.

There is a full dataset with a realistic density here:
`/data/users/Public/couchard/world_cereal/2019_BE_Flanders_rdm_grid_samples.gpkg`

So far, I managed to run jobs with 100 samples, with a peak memory usage a bit higher than 5GB, and 3hours of processing costing 400 credits (CDSE-STAGING: j-2403205396124e2f8d417a0c39667ced). Memory usage seems fine but it would be nice to speed-up the job.

I did other tests with much less samples and it seems to me that the memory scales slowly up with the number of samples, which seems good.


This line here subsamples the divided openeo jobs into a single one for testing purposes:
https://github.com/WorldCereal/worldcereal-classification/blob/9a14cee8cf0e5f414841fd94ce811e71acdb74d6/scripts/extractions/extract_optical.py#L332

GriffinBabe:
@jdries 
According our last discussion I generated a new job which is currently running
- A conservative mask is applied
- The SCL Dilated masks and the distance to clouds are both computed from the same input collection
- I removed useless resample_cube_spatial and resample_spatial operations
- I set the memory to 6GB (executor memory + overhead) instead of 8GB
- Now sampling 250 locations, but will try increasing this later to 500 or 1000

GriffinBabe:
S1 pipeline seems to work well, managed to extract 500 samples with close to 6GB of memory

jdries:
Found another cause of memory usage in S2: the distance to cloud computation works on the whole timeseries, while it should only need individual observations.
So the size argument needs to include time dimension:
```
{
      "dimension": "t",
     "unit": null,
     "value": "P1D"
}
```

This will then also require the UDF to be adjusted, maybe need to check with @VincentVerelst

jdries:
Another one: the cloud mask is now applied, but there's a 'merge_cubes' before the mask process.
I think this will still cause a full data load, which we try to avoid. Switching masking and merging should solve this. 
There's then probably still the question of avoiding dates for masked data in the output netcdf. Need to see how to solve that, perhaps by setting the fully masked areas to nodata.

GriffinBabe:
@jdries Without the two optimizations you mentionned here I managed to make the extraction of 500 points for 8GB of peak memory, 580 credits and 4hours wall time (cdse-staging j-24032149eeb847fbbfdc5a6bfca9622c). I will run a new one with the same amount of samples and see the difference in memory usage.

I also ran an extraction on S1 data with 6GB peak memory for 500 points, 160 credits and ~1h30 wall time. Seems quite efficient to me, but could you give it a look to see how can we optimize it? (cdse production j-240321d24a59494d848ae4eedc695edf)

jdries:
For sentinel-1, can you try setting executor-memory to 1G, keeping memory-overhead the same.

jdries:
For sentinel-2, I'm going to try writing an alternative loader. May take a little bit of time, but might be worth it as I again notice the big difference between loading data in bulk vs samples.

jdries:
New strategy looks promising, job ran with 2.5GB per executor, taking only 57 cpu hours in spark.
j-240323df984c4f4dac7661f80021c135
Settings:
```
{
      "driver-memory": "2G",
      "driver-memoryOverhead": "2G",
      "driver-cores": "1",
      "executor-memory": "600m",
      "executor-memoryOverhead": "1900m",
      "executor-cores": "1",
      "max-executors": "80",
      "soft-errors": "true",
      "gdal-dataset-cache-size":2,
      "gdal-cachemax":120,
      "executor-threads-jvm":1
  }
```
Still need to work on ensuring correctness of the result. Also the end of the job takes too long to look up sample metadata. This could probably be done in parallel as a quickfix.
Now that the actual reading is a lot faster, there's also other areas to look at.

GriffinBabe:
@jdries For the S1 job, I think that limiting the memory overhead of the executor to 1GB is too little. This jobs seems to crash (I think for OOM reasons)

j-240325d99399443cabd88a9b4c0fb17f CDSE-STAGING

jdries:
@GriffinBabe  my suggestion was:
> try setting executor-memory to 1G, keeping memory-overhead the same
you tried the other way round, which I would indeed expect to crash.
executor-memory is used by java and spark, which is fairly good at managing memory.
memory-overhead is for python and C code like orfeo and gdal, which is quite bad at keeping memory within bounds, hence jobs often need less executor-memory, and more overhead...

jdries:
A working version of the performance improvement is available on CDSE staging. Patches look fine upon first inspection, but deeper comparison is needed. To enable the optimization, we need to set a feature flag on the load_collection call that loads the optical bands. Here's an example of how to do that (the second line):

```
cube = connection.load_collection("SENTINEL2_L2A",bands=["B02","B03","B04"],temporal_extent=["2023-08-01","2023-08-31"],max_cloud_cover=85)
    cube.result_node().update_arguments(featureflags={"tilesize": 128, "load_per_product": True})
```

Also the job options have a large impact on performance. Note that I run with a small number of executors, to minimize loss when idle, but you can increase it if you want to iterate faster:

    ```
job_options = {
        "driver-memory": "2G",
        "driver-memoryOverhead": "2G",
        "driver-cores": "1",
        "executor-memory": "600m",
        "executor-memoryOverhead": "1900m",
        "executor-cores": "1",
        "max-executors": "26",
        "soft-errors": "true",
        "gdal-dataset-cache-size":2,
        "gdal-cachemax":120,
        "executor-threads-jvm":1
    }
```

Note that the feature flag enables a very experimental code path, please don't start using it everywhere without asking or properly checking results. Processes like mask_scl_dilation are _not_ yet supported for sure.

As for next steps, I would recommend to start up sample extraction where we do not yet apply cloud masking, and then properly verify results. Then we can add the cloud masking when the more complex strategy works.

I also still see other things that can be made faster, but takes some time to integrate everything.

File: scripts/extractions/extract_meteo.py [hunk 1]
```diff
@@ -0,0 +1,162 @@
+"""Extract AGERA5 (Meteo) data using OpenEO-GFMAP package."""
+import argparse
+from functools import partial
+from pathlib import Path
+
+import geojson
+import geopandas as gpd
+import openeo
+import pandas as pd
+from extract_sar import (
```
kvantricht, line 10: Is this the most elegant way to do this? Importing private methods from other script seems wild. Maybe we should have this integrated as non-private methods in gfmap or have some utilities here that exposes them.

scripts/extractions/extract_meteo.py
```diff
@@ -0,0 +1,162 @@
...
+    _pipeline_log,
+    _setup_logger,
+    _upload_geoparquet_artifactory,
+    create_job_dataframe,
+    generate_output_path,
+)
+from openeo_gfmap import Backend, TemporalContext
+from openeo_gfmap.backend import vito_connection
+from openeo_gfmap.manager.job_manager import GFMAPJobManager
+from openeo_gfmap.manager.job_splitters import (
```
kvantricht, line 22: also here importing private methods. Can we avoid this?

scripts/extractions/extract_meteo.py
```diff
@@ -0,0 +1,162 @@
...
+
+    # Filter the geometry to the rows with the extract only flag
+    geometry = _filter_extract_true(geometry)
+    assert len(geometry.features) > 0, "No geometries with the extract flag found"
+
+    # Performs a buffer of 64 px around the geometry
+    geometry_df = _buffer_geometry(geometry, distance_m=5)
+    spatial_extent_url = _upload_geoparquet_artifactory(geometry_df, row.name)
+
+    bands_to_download = ["temperature-mean"]
```
kvantricht, line 53: we also need the precipitation!

scripts/extractions/extract_meteo.py
```diff
@@ -0,0 +1,162 @@
...
+
+    cube = connection.load_collection(
```
kvantricht, line 55: are we doing a rescaling of meteo variables somewhere? We should convert to integers to save storage (I can point you at default scaling)

scripts/extractions/extract_optical.py
```diff
@@ -0,0 +1,380 @@
...
+from pathlib import Path
+from typing import List
+
+import geojson
+import geopandas as gpd
+import openeo
+import pandas as pd
+import pystac
+import xarray as xr
+from extract_sar import (
```
kvantricht, line 16: same comments on private methods

scripts/extractions/extract_sar.py
```diff
@@ -0,0 +1,400 @@
...
+    class ManagerLoggerFilter(logging.Filter):
+        """Filter to only accept the OpenEO-GFMAP manager logs."""
+
+        def filter(self, record):
+            return record.name in [_log.name, _pipeline_log.name]
+
+    stream_handler.addFilter(ManagerLoggerFilter())
+
+
+def _buffer_geometry(
```
kvantricht, line 62: why is this not part of GFMAP?

scripts/extractions/extract_sar.py
```diff
@@ -0,0 +1,400 @@
...
+    split_job_s2grid,
+)
+from openeo_gfmap.stac import AUXILIARY
+from shapely.geometry import Point
+
+# Logger for this current pipeline
+_pipeline_log: Optional[logging.Logger] = None
+
+
+def _setup_logger(level=logging.INFO) -> None:
```
kvantricht, line 37: Why is this not part of GFMAP?

kvantricht (commented):
Nice work, though I have to say it's quite lengthy code and it doesn't feel very modular. If this is what a user typically needs to do, it's quite something. I have the feeling that some of it could be embedded in GFMAP and also be written a bit more modular, so the actual user-based extraction scripts are also more readable. What do you think?

GriffinBabe:
@kvantricht This pull request will be much larger than it used to be, as I also started working on the inference blocks using GFMAP to extract the data in WorldCereal, instead of the usual recycled cropclass/HRL code that was in the repository. 

The old code will still be accessible through the versioning, but I removed it in this branch (to be merged whenever we confirm that a fetched datacube will have the same values here as it had with the previous pipeline)

kvantricht:
> @kvantricht This pull request will be much larger than it used to be, as I also started working on the inference blocks using GFMAP to extract the data in WorldCereal, instead of the usual recycled cropclass/HRL code that was in the repository.
> 
> The old code will still be accessible through the versioning, but I removed it in this branch (to be merged whenever we confirm that a fetched datacube will have the same values here as it had with the previous pipeline)

Did you tackle my previous comments? Wouldn't it be good to have a new stable version (so a merged PR) before you're actually gone?

scripts/extractions/extract_meteo.py
GriffinBabe: I changed them to non-private methods, except for some in GFMAP

scripts/extractions/extract_meteo.py
GriffinBabe: @kvantricht can you point me at that scaling?

scripts/extractions/extract_sar.py
GriffinBabe: `_buffer_geometry` here is related to the use-case. We can have an implementation in GFMAP but I would wait for the implementation of the pipeline from the data engineering team

scripts/extractions/extract_sar.py
GriffinBabe: This `_setup_logger ` function setups print in console for the logger objects in GFMAP Manager, from the extraction script and open-eo

I moved to GFMAP the setup for it's own logger, but I will leave a function here to setup the logger object of the extraction pipeline

scripts/extractions/extract_meteo.py
GriffinBabe: This is done

GriffinBabe:
@kvantricht I resolved your comments and updates the pipelines. I think there's going to be a few changes later when extracting S1 and METEO but we have a solid foundation

scripts/extractions/extract_meteo.py
```diff
@@ -0,0 +1,166 @@
...
+    filter_geometry = connection.load_url(spatial_extent_url, format="parquet")
+    cube = cube.filter_spatial(filter_geometry)
+    cube.rename_labels(
+        dimension="bands",
+        target=["AGERA5-temperature-mean", "AGERA5-precipitation-flux"],
+        source=["temperature-mean", "precipitation-flux"],
+    )
+
+    # Rescale to uint16, multiplying by 100 first
+    cube = cube * 100
```
kvantricht, line 69: So it is confirmed that `cube` contains the unscaled values which is why you reintroduce it here?

scripts/extractions/extract_optical.py
```diff
@@ -0,0 +1,380 @@
...
+
+# Define the sentinel 2 asset
+sentinel2_asset = pystac.extensions.item_assets.AssetDefinition(
+    {
+        "gsd": 10,
+        "title": "Sentinel2",
+        "description": "Sentinel-2 bands",
+        "type": "application/x-netcdf",
+        "roles": ["data"],
+        "proj:shape": [64, 64],
```
kvantricht, line 42: Does this have to be hardcoded here? What if patch size changes?

kvantricht (approved):
Approved for now after discussion with Darius. Some future changes planned.

PR #48, opened 2024-05-31
44 translate cropno crop inference udf to gfmap

GriffinBabe:
FYI: The branch `hv_mvp_inferenceUDF` is already merged here

About the folder `minimal_wc_presto`: I deleted all the duplicates UDFs. I suppose that want to keep the dependency code somewhere, but should we leave this here?

I added the Feature Extractor and Model Inference classes to the module's source, as well as updated a script to perform cropland mapping from coordinates provided in the command lines.

The UDFs are now using all the exporter dependencies packed by @HansVRP and are using GFMAP for fetching the data from the collections, resulting in less (and hopefully for you, more readable) code.

I also took the liberty in deleting an old Feature Extractor UDF (I suppose we don't need it anymore)


minimal_wc_presto/ONNX_conversion.py
kvantricht, line 1: Is it in general possible to "get rid" of the `minimal_wc_presto` folder? I think it was meant to live temporarily to contain Hans' experiments to make the UDF work. Now that conversion is done, I would try to remove this folder and add any required stuff that's still in there to a more logical place.

scripts/inference/cropland_mapping.py
```diff
@@ -1,21 +1,21 @@
...
-from openeo_gfmap.backend import Backend, BackendContext, cdse_connection
-from openeo_gfmap.features.feature_extractor import PatchFeatureExtractor
+from openeo_gfmap.backend import Backend, BackendContext
+from openeo_gfmap.features.feature_extractor import apply_feature_extractor
+from openeo_gfmap.inference.model_inference import apply_model_inference
 
+from worldcereal.openeo.feature_extractor import PrestoFeatureExtractor
+from worldcereal.openeo.inference import CroplandClassifier
 from worldcereal.openeo.preprocessing import worldcereal_preprocessed_inputs_gfmap
+from worldcereal.openeo.feature_extractor import PrestoFeatureExtractor
```
kvantricht, line 15: double import?

scripts/inference/cropland_mapping.py
```diff
@@ -1,21 +1,21 @@
...
 
-
-class PrestoFeatureExtractor(PatchFeatureExtractor):
-    def __init__(self):
-        pass
-
-    def extract(self, image):
-        pass
-
+ONNX_DEPS_URL = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/openeo/onnx_dependencies_1.16.3.zip"
```
kvantricht, line 18: does this also include pytorch? In this case, is this name transparent enough?

scripts/inference/cropland_mapping.py
```diff
@@ -41,29 +50,70 @@ def extract(self, image):
...
     )
 
-    # Start the job and download
-    job = input_cube.create_job(
-        title=f"Cropland inference BBOX: {minx} {miny} {maxx} {maxy}",
-        description="Cropland inference using WorldCereal, Presto and GFMAP classifiers",
-        out_format="NetCDF",
+    # Test feature computer
+    presto_parameters = {
+        "rescale_s1": False,  # Will be done in the Presto UDF itself!
```
kvantricht, line 77: A little confusing, let's discuss this. If needed, we can also adapt the Presto UDF and remain consistent.

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+    """Feature extractor to use Presto model to compute embeddings.
+    This will generate a datacube with 128 bands, each band representing a
+    feature from the Presto model.
+    """
+
+    import functools
+    from pathlib import Path
+    from typing import Tuple
+
+    PRESTO_PATH = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal-minimal-inference/presto.pt"  # NOQA
```
kvantricht, line 16: I would put this as optional or even required argument when creating an instance of this class, as there will be multiple (finetuned) Presto model weights. The current one is finetuned for crop/no-crop, and can (for now) be the default, but still we need the option to provide other weights.

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+    BASE_URL = "https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies"  # NOQA
```
kvantricht, line 17: `project_dependencies`doesn't say much. And how is this now different from the ONNX dependencies?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+    DEPENDENCY_NAME = "wc_presto_onnx_dependencies.zip"
```
kvantricht, line 18: Yet another dependencies zip. This is getting confusing. Is there a way to cover all needs into one dependencies file setup specifically for WorldCereal?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+        "S2-L2A-B06": "B06",
+        "S2-L2A-B07": "B07",
+        "S2-L2A-B08": "B08",
+        "S2-L2A-B8A": "B8A",
+        "S2-L2A-B11": "B11",
+        "S2-L2A-B12": "B12",
+        "S1-SIGMA0-VH": "VH",
+        "S1-SIGMA0-VV": "VV",
+        "COP-DEM": "DEM",
+        "A5-tmean": "temperature-mean",
```
kvantricht, line 34: I think we haven't discuss these band names yet. I get the temptation of sticking always to 2 character 'sensor' specification but `A5` really doesn't say much and often in the future it won't be possible in a good way to stick to two characters anyway. I propose to use the full `AGERA5` specification in the GFMAP band names. Shouldn't we also stick to uppercase? `TMEAN` in that case?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+        "A5-precip": "precipitation-flux",
```
kvantricht, line 35: I propose: `AGERA5-PRECIP`

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+    }
+
+    def __init__(self):
+        """
+        Initializes the PrestoFeatureExtractor object, starting a logger.
```
kvantricht, line 40: Isn't starting a logger something GFMAP can take care of, without the use having to specify this part himself? In that case this entire `init` could be inherited from parent class, except for the fact that probably here we need to add the optional presto path argument.

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+        """
+        import logging
+
+        logging.basicConfig(level=logging.INFO)
+        self.logger = logging.getLogger(PrestoFeatureExtractor.__name__)
+
+    @classmethod
+    @functools.lru_cache(maxsize=6)
+    def extract_dependencies(cls, base_url: str, dependency_name: str):
```
kvantricht, line 49: Not an expert here, but just wondering if this is not something that could be defined inside GFMAP instead of here? That IF custom dependencies are provided, these are automatically extracted?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+        shutil.unpack_archive(modelfile, extract_dir=dependencies_dir)
+
+        # Add the model directory to system path if it's not already there
+        abs_path = str(
+            dependencies_dir / Path(modelfile_url).name.split(".zip")[0]
+        )  # NOQA
+
+        return abs_path
+
+    def output_labels(self) -> list:
```
kvantricht, line 77: Have you tested if these are in fact attached to the dataset? I've seen in the past that band names altered in the UDF are not picked up by OpenEO unless specifically using an OpenEO rename.

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,115 @@
...
+        inarr = inarr.fillna(65535)
+
+        # Unzip de dependencies on the backend
+        self.logger.info("Unzipping dependencies")
+        deps_dir = self.extract_dependencies(self.BASE_URL, self.DEPENDENCY_NAME)
+
+        self.logger.info("Appending dependencies")
+        sys.path.append(str(deps_dir))
+
+        from dependencies.wc_presto_onnx_dependencies.mvp_wc_presto.world_cereal_inference import get_presto_features
```
kvantricht, line 111: To be discussed with @HansVRP I would restructure the dependencies archive and avoid the `mvp_wc_presto` part which should not be part of the official dependencies.

src/worldcereal/openeo/inference.py
```diff
@@ -0,0 +1,67 @@
+"""Model inference on Presto feature for binary classication"""
+
+import xarray as xr
+from openeo_gfmap.inference.model_inference import ModelInference
+
+
+class CroplandClassifier(ModelInference):
+    import functools
+
+    CATBOOST_PATH = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal-minimal-inference/wc_catboost.onnx"  # NOQA
```
kvantricht, line 10: Same remark as with the feature extractor. This is the default model path, but we have to be able to set another one when creating an instance of this class.

src/worldcereal/openeo/inference.py
```diff
@@ -0,0 +1,67 @@
...
+    DEPENDENCY_NAME = "wc_presto_onnx_dependencies.zip"
+
+    def __init__(self):
+        import logging
+        logging.basicConfig(level=logging.INFO)
+        self.logger = logging.getLogger(WorldCerealInference.__name__)
+
+    @classmethod
+    @functools.lru_cache(maxsize=6)
+    def extract_dependencies(cls, base_url: str, dependency_name: str):
```
kvantricht, line 21: Same comment as before, can't we take care of this in GFMAP and shorten the project-specific class?

src/worldcereal/openeo/preprocessing.py
```diff
@@ -193,6 +193,22 @@ def raw_datacube_DEM(
...
+def raw_datacube_METEO(
+    connection: Connection,
+    backend_context: BackendContext,
+    spatial_extent: SpatialContext,
+    temporal_extent: TemporalContext,
+    fetch_type: FetchType,
+) -> DataCube:
+    extractor = build_generic_extractor(
+        backend_context=backend_context,
+        bands=["A5-tmean", "A5-precip"],
```
kvantricht, line 205: Commented earlier, need to discuss about band names.

src/worldcereal/openeo/preprocessing.py
```diff
@@ -236,16 +250,15 @@ def worldcereal_preprocessed_inputs_gfmap(
         spatial_extent=spatial_extent,
         temporal_extent=temporal_extent,
         bands=[
-            "S1-SIGMA0-VV",
             "S1-SIGMA0-VH",
+            "S1-SIGMA0-VV",
         ],
         fetch_type=FetchType.TILE,
         target_resolution=10.0,  # Compute the backscatter at 20m resolution, then upsample nearest neighbor when merging cubes
         orbit_direction="ASCENDING",
```
kvantricht, line 258: This is artefact from before. We shouldn't be hardcoding this but choose an orbit dynamically based on availability of observations.

minimal_wc_presto/ONNX_conversion.py
GriffinBabe: Yes, I agree
To discuss at the GFMAP meeting

src/worldcereal/openeo/feature_extractor.py
GriffinBabe: I added a "presto_url" parameter

src/worldcereal/openeo/feature_extractor.py
GriffinBabe: This is much more than ONNX as it contains pytorch and all the other presto dependencies

src/worldcereal/openeo/feature_extractor.py
GriffinBabe: That's the idea of it. The Feature Computer is initialized first locally in order to call this `output_labels` function, which then will be added on the OpenEO computing graph

kvantricht:
@GriffinBabe I pushed a new Presto wheel to Artifactory and added what is supposed to become the new way of using dependency wheel in the UDF in my commit. Once this works, we may even not need the dependency zip in this UDF anymore (to be verified). Good luck testing the new functionality. Let me know if you need anything from my side.

HansVRP:
I am also rehashing over the UDF trying to load the required libraries from within the class initiation. I already managed to pull the catboost predictor definition out of the UDF_long. I will try to do the same for the Presto feature extractor. 

If it works we might not to upload mvp_world_cereal inference anywhere

kvantricht:
> I am also rehashing over the UDF trying to load the required libraries from within the class initiation. I already managed to pull the catboost predictor definition out of the UDF_long. I will try to do the same for the Presto feature extractor.
> 
> If it works we might not to upload mvp_world_cereal inference anywhere

This may be parallel work with what I’m doing Hans, as I discussed with Darius and migrated all that functionality to presto-WorldCereal library where it belongs. We want to test the new functionality defining an external whl on the top of the UDF.

HansVRP:
> > I am also rehashing over the UDF trying to load the required libraries from within the class initiation. I already managed to pull the catboost predictor definition out of the UDF_long. I will try to do the same for the Presto feature extractor.
> > If it works we might not to upload mvp_world_cereal inference anywhere
> 
> This may be parallel work with what I’m doing Hans, as I discussed with Darius and migrated all that functionality to presto-WorldCereal library where it belongs. We want to test the new functionality defining an external whl on the top of the UDF.

Did PrestoFeatureExtractor also end up in the whl? It is indeed not required!

kvantricht:
> > > I am also rehashing over the UDF trying to load the required libraries from within the class initiation. I already managed to pull the catboost predictor definition out of the UDF_long. I will try to do the same for the Presto feature extractor.
> > > If it works we might not to upload mvp_world_cereal inference anywhere
> > 
> > 
> > This may be parallel work with what I’m doing Hans, as I discussed with Darius and migrated all that functionality to presto-WorldCereal library where it belongs. We want to test the new functionality defining an external whl on the top of the UDF.
> 
> Did PrestoFeatureExtractor also end up in the whl? It is indeed not required!

Presto-related stuff now went here: https://github.com/WorldCereal/presto-worldcereal/blob/openeo_inference/presto/inference.py

scripts/inference/cropland_mapping.py
```diff
@@ -1,21 +1,19 @@
...
 
-
-class PrestoFeatureExtractor(PatchFeatureExtractor):
-    def __init__(self):
-        pass
-
-    def extract(self, image):
-        pass
-
+ONNX_DEPS_URL = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/openeo/onnx_dependencies_1.16.3.zip"
```
kvantricht, line 16: Why is this still needed? `onnxruntime` is included in the `worldcereal_deps.zip`. And if it's still needed, shouldn't we also put it on s3?

scripts/inference/cropland_mapping.py
```diff
@@ -27,12 +25,20 @@ def extract(self, image):
     parser.add_argument("miny", type=float, help="Minimum Y coordinate (south)")
     parser.add_argument("maxx", type=float, help="Maximum X coordinate (east)")
     parser.add_argument("maxy", type=float, help="Maximum Y coordinate (north)")
+    parser.add_argument(
+        "--epsg",
```
kvantricht, line 29: What does this argument do exactly? Is this for output EPSG? Internally there is still the datacube epsg which is used to feed lat/lon to Presto?

scripts/inference/cropland_mapping.py
```diff
@@ -41,29 +47,70 @@ def extract(self, image):
...
     )
 
-    job.start_and_wait()
-    job.get_results().download_files(args.output_folder)
+    classes.execute_batch(
+        outputfile=args.output_path,
+        out_format="NetCDF",
+        job_options={
+            "driver-memory": "4g",
+            "executor-memoryOverhead": "12g",
```
kvantricht, line 113: That's very high. Need to flag this for further profiling. But for now we can leave it like this.

scripts/inference/cropland_mapping.py
```diff
@@ -41,29 +47,70 @@ def extract(self, image):
...
+            "udf-dependency-archives": [f"{ONNX_DEPS_URL}#onnx_deps"],
```
kvantricht, line 114: cfr earlier comment. Why still needed?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -0,0 +1,111 @@
...
+      dependencies already take care of the backscatter decompression. If
+      specified, should be set as `False`.
+    """
+
+    PRESTO_MODEL_URL = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal-minimal-inference/presto.pt"  # NOQA
+    PRESO_WHL_URL = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/dependencies/presto_worldcereal-0.1.0-temp-py3-none-any.whl"
+    BASE_URL = "https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies"  # NOQA
+    DEPENDENCY_NAME = "worldcereal_deps.zip"
+
+    GFMAP_BAND_MAPPING = {
```
kvantricht, line 26: This is in fact a bit an overhead, if later on in `presto-worldcereal` we again remap: https://github.com/WorldCereal/presto-worldcereal/blob/main/presto/inference.py#L41

So we should consider to immediately use the presto naming here and remove the remapping in `presto-worldcereal`

scripts/inference/cropland_mapping.py
GriffinBabe: Because it's much lighter than all the presto dependency so it's a waste of resources to install everything just for `onnxruntime`. Also it's a requirement in GFMAP for the ModelInference subclasses, until the pip dependency feature is more stable.

Indeed, could you updload it to the S3 @HansVRP ?

scripts/inference/cropland_mapping.py
GriffinBabe: No this is the epsg for the input bounds minx, miny, maxx, maxy

I modified the description to the parameter to clarify this

scripts/inference/cropland_mapping.py
kvantricht: Does this mean that for each single UDF in the chain those dependencies get downloaded/unpacked again? That's also lots of overhead I guess?

GriffinBabe: In this case it's just onnxruntime that get's reinstalled, compared to the full UDF I don't think there is a large difference in execution time, even less with large spatial extents probably

kvantricht (approved):

scripts/inference/cropland_mapping.py
HansVRP: From what I have seen it is a potential overkill, 200 km square ran on half the amount of memory for me. @GriffinBabe do you have a job id for a job which needed more memory?

PR #59, opened 2024-06-12
Make `batch_size` for Presto flexible

kvantricht:
`presto-worldcereal` wheel was updated and `get_presto_features` now accepts an optional `batch_size` parameter. In this PR, it's now set to 4096 (but not yet tested). Idea is that we can see what works best for our case.


GriffinBabe (approved):

HansVRP:
It might be convenient to couple the batch and chuck size to each other per standard value. 

If not I would go from an apply_datacube to a reduce dimension. 

Being able to both control the batch and chunk size can lead to less optimized inference workflows for less experienced users.

Issue #62, opened 2024-06-19
Generate GeoTIFF outputs instead of NetCDF


kvantricht:
Jobs now generate NetCDF outputs. We'd like that to be GeoTIFF outputs.

GriffinBabe:
Closed by #65

PR #65, opened 2024-06-19
Now writing GeoTiff and having two bands: binary mask and probabilities

### LINKED ISSUE #61: Return probabilities from inference job
At the moment we return a binarized classification product from the inference job. At the very least we should allow to return the probabilities. For now we can deal with that in `worldcereal-classification`, later on we probably need a postprocessing job defined in `gfmap` as well.

GriffinBabe:
Resolves #61 and #62 

Output GeoTiff example for parameters `664000 5611120 665000 5612120 2020-11-01 2021-10-31 --epsg 32631` can be found in my Public drive:
`/data/users/Public/couchard/gfmap_cropland_inference_probabilities.tif`

Output GeoTiff has two bands, the first one is the binary cropland mask and the second is the associated probabilities

### DIFF
--- a/scripts/inference/cropland_mapping.py
+++ b/scripts/inference/cropland_mapping.py
@@ -8,6 +8,7 @@ from openeo_gfmap import BoundingBoxExtent, TemporalContext
 from openeo_gfmap.backend import Backend, BackendContext
 from openeo_gfmap.features.feature_extractor import apply_feature_extractor
 from openeo_gfmap.inference.model_inference import apply_model_inference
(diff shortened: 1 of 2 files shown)


kvantricht (changes requested):
Could we scale probabilities by a factor 100 to be able to save the result as integer instead of float? Need to see what kind of integer given that the label for now is binary but in croptype can become quite large.

GriffinBabe:
@kvantricht I updated this to have uint8 values with 0-1 for crop/no-crop first layer and 0-100 for the probability layer.

kvantricht (approved):

PR #84, opened 2024-06-26
Point extractions

VincentVerelst:
@jdegerickx, you can find the output of the point extractions script here: _/data/users/Public/vincent.verelst/gfmap_feature_extractions/35NPF/2018_SSD_WFP-field-survey_POLY_110_1674.parquet_


VincentVerelst:
@jdegerickx , PR is ready for review

scripts/extractions/point_extractions/point_extractions.py
```diff
@@ -0,0 +1,280 @@
...
+    class ManagerLoggerFilter(logging.Filter):
+        """Filter to only accept the OpenEO-GFMAP manager logs."""
+
+        def filter(self, record):
+            return record.name in [pipeline_log.name]
+
+    stream_handler.addFilter(ManagerLoggerFilter())
+
+
+def filter_extract_true(
```
jdegerickx, line 47: the "extract" attribute is an integer which is not just binary in nature.
In any case, a value of 0 means no extractions should be done for the sample.
Numbers other than zero indicate the priority for extractions and value "99" means extractions have already been done in Phase I. 
So instead of selecting all samples having value 1, I would drop samples having value 0

scripts/extractions/point_extractions/point_extractions.py
```diff
@@ -0,0 +1,280 @@
...
+) -> list:
+    for idx, item in enumerate(job_items):
+        item_asset_path = Path(list(item.assets.values())[0].href)
+
+        gdf = gpd.read_parquet(item_asset_path)
+
+        # Convert the dates to datetime format
+        gdf["date"] = pd.to_datetime(gdf["date"])
+
+        # Convert band dtype to uint16 (temporary fix)
```
jdegerickx, line 186: guess this won't work if you decide to only extract S2?

scripts/extractions/point_extractions/point_extractions.py
```diff
@@ -0,0 +1,280 @@
...
+    tracking_df_path = Path(args.output_path) / "job_tracking.csv"
+
+    # Load the input dataframe, and perform dataset splitting using the h3 tile
+    # to respect the area of interest. Also filters out the jobs that have
+    # no location with the extract=True flag.
+    pipeline_log.info("Loading input dataframe from %s.", args.input_df)
+
+    input_df = gpd.read_parquet(args.input_df)
+
+    split_dfs = split_job_s2grid(input_df, max_points=args.max_locations)
```
jdegerickx, line 252: we have an ongoing discussion in teams (https://teams.microsoft.com/l/message/19:b2237c00ef4d4b0da769872d202fdb32@thread.tacv2/1719393187427?tenantId=9e2777ed-8237-4ab9-9278-2c144d6f6da3&groupId=d6f94f76-49f4-4ce2-9c34-c86de888fed5&parentMessageId=1719393187427&teamName=WorldCereal%20Phase%202&channelName=VITO_internal&createdTime=1719393187427&ngc=true) on how to deal with datasets showing a large variety in validity times.
As a temporary solution, could you add an additional splitting based on valid_time?
For now I would use a threshold of 4 months: so you should group all samples together with valid_times closer than 4 months from each other into a single job. See for instance:
https://git.vito.be/projects/APPL/repos/cropclass/browse/src/cropclass/trainingdata/datasets.py#296-303

src/worldcereal/openeo/preprocessing.py
```diff
@@ -276,7 +280,17 @@ def worldcereal_preprocessed_inputs_gfmap(
     backend_context: BackendContext,
     spatial_extent: BoundingBoxExtent,
     temporal_extent: TemporalContext,
+    fetch_type: Optional[FetchType] = FetchType.TILE,
+    collections: Optional[List[str]] = None,
 ) -> DataCube:
+    if not all(
+        coll in {"sentinel1", "sentinel2", "copernicus", "meteo"}
```
jdegerickx, line 287: please use "cop_dem" or "dem" as collection name instead of "copernicus"

src/worldcereal/openeo/preprocessing.py
jdegerickx, line 1: would be nice to have an additional optional argument "settings", specifying:
- which bands to extract from each collection
- compositing method to use (for now only median compositing supported, but we would like to test BAP compositing as well in the near future)
- the compositing window --> I typically use "month" and "dekad"

jdegerickx (changes requested):
See individual comments posted in the changed files.

scripts/extractions/point_extractions/point_extractions.py
VincentVerelst: Indeed you will have to manually adjust the bands if you decide to only extract a certain collection. The converting to uint16 is done in the post job actions as a temporary solution. Normally, it's done on the openeo backend, but there is a bug there: https://github.com/Open-EO/openeo-geopyspark-driver/issues/805

src/worldcereal/openeo/preprocessing.py
VincentVerelst: Resolved by latest commit

scripts/extractions/point_extractions/point_extractions.py
VincentVerelst: Linking the open issue for job splitting on temporal intervals in GFMap: https://github.com/Open-EO/openeo-gfmap/issues/79

scripts/extractions/point_extractions/point_extractions.py
VincentVerelst: Ok, will adjust. Important to note that we do use the filtering on only value 1 also for the patch extractions: https://github.com/WorldCereal/worldcereal-classification/blob/main/scripts/extractions/extract_sar.py#L84

VincentVerelst: Resolved by latest commit

src/worldcereal/openeo/preprocessing.py
VincentVerelst: Maybe we can open a separate branch for this? Especially the BAP compositing will take some time as we have ongoing discussions with the openeo devs on how to best support BAP in openeo (maybe make a public UDP out of it)

jdegerickx: I agree we should not include BAP compositing right now, but perhaps already make sure these things can be easily added in the future? So instead of "collections" being a list, could we make it a dict in which you can specify some additional settings for each collection? Or is this not the right way to pass these settings?

jdegerickx: agreed to tackle this at a later stage

jdegerickx (approved):

PR #85, opened 2024-06-27
Update crop type mapping workflow

kvantricht:
This workflow is still the modular one with separated UDFs for feature extraction and model inference as for the moment there is no indication that merging them speeds things up. Ultimately we still may want to do the merging to have a cleaner process graph that is easier to debug and optimize.

This PR includes automated cropland masking for the crop type products, including @jdries verified temporary fix of `reduce_dimension` on the cropland mask. By running `generate_map` with `product_type="croptype"`, the crop type workflow can be triggered.

Example usage (both for cropland and croptype) can be found in `scripts/inference/cropland_mapping.py`.

Currently specifying a custom model URL is not supported yet. Need to discuss with @GriffinBabe on an elegant way to parse parameters.

### DIFF
--- a/src/worldcereal/job.py
+++ b/src/worldcereal/job.py
@@ -1,10 +1,12 @@
 """Executing inference jobs on the OpenEO backend."""
+
 from dataclasses import dataclass
 from enum import Enum
 from pathlib import Path
 from typing import Optional, Union
 
 import openeo
+from openeo import DataCube
 from openeo_gfmap import BackendContext, BoundingBoxExtent, TemporalContext
 from openeo_gfmap.features.feature_extractor import apply_feature_extractor
 from openeo_gfmap.inference.model_inference import apply_model_inference
@@ -87,6 +89,7 @@ def generate_map(
     output_path: Optional[Union[Path, str]],
     product_type: WorldCerealProduct = WorldCerealProduct.CROPLAND,
     out_format: str = "GTiff",
+    apply_cropland_mask: bool = False,
 ):
     """Main function to generate a WorldCereal product.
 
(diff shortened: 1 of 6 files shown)


src/worldcereal/job.py
```diff
@@ -100,80 +102,65 @@ def generate_map(
 
     Raises:
         ValueError: if the product is not supported
+        ValueError: if the out_format is not supported
```
GriffinBabe, line 105: Needs to be discussed but in the preprocessing.py file we are using a different docstyle (numpy)

Ideally we should choose a unique doc style for the whole project. GFMAP is also using the numpy docstyle

For example:
https://github.com/WorldCereal/worldcereal-classification/blob/main/src/worldcereal/openeo/preprocessing.py#L35

src/worldcereal/job.py
```diff
@@ -218,3 +205,120 @@ def collect_inputs(
...
+            {"dimension": "y", "unit": "px", "value": 0},
+        ],
+    )
+
+    # Mask cropland
+    if cropland_mask is not None:
+        classes = classes.mask(cropland_mask == 0, replacement=0)
+
+    # Cast to uint16
+    classes = compress_uint16(classes)
```
GriffinBabe, line 322: This is something that needs to be changed to uint32 right?

@jdries Is there a data type change trigger if I do, let's say
`cube = cube.linear_scale_range(0, 4_294_967_295 - 1, 0, 4_294_967_295 - 1)` ?

GriffinBabe: (`4_294_967_295` is the uint32 max value)

src/worldcereal/job.py
```diff
@@ -218,3 +205,120 @@ def collect_inputs(
         out_format="NetCDF",
         job_options={"driver-memory": "4g", "executor-memoryOverhead": "4g"},
     )
+
+
+def _cropland_map(inputs: DataCube) -> DataCube:
```
jdegerickx, line 210: probably this function and the _croptype_map function can be combined into one inference function to avoid duplication? Just adding an additional "product" parameter might do the trick?
But not essential for now I'd say...

jdegerickx (commented):
Only one comment from my side but it's not a breaking issue at the moment so we can proceed

src/worldcereal/job.py
kvantricht: ah good point, I just use the autodocstring extension in vscode which produces this stuff. Should check if I can configure that somehow.

src/worldcereal/job.py
kvantricht: at the moment we're still fine as we just have few classes and simple integers representing them but I agree this needs to be resolved when we move to other legend.

src/worldcereal/job.py
kvantricht: Apparently it was indeed a setting in the extension. Updated the docstrings to numpy style.

GriffinBabe (approved):

PR #91, opened 2024-07-02
Added configuration data classes for cropland/croptype mapping workflow

GriffinBabe:
@kvantricht I added data classes for parameters that use `pydantic`, a very popular module to enforce the data types as described in the type hints (the module was already included in worldcereal, probably as a subdependency so it doesn't make our package more heavy).

One way of using them would be:
```python
from worldcereal.job import generate_map, WorldCerealProduct, CropTypeParameters

croptype_parameters = CropTypeParameters()
croptype_parameters.classifier_parameters.classifier_url = "https://s3.amazon.com/example/link.onnx"

inference_results = generate_map(
    spatial_extent=...,
    temporal_extent=...,
    backend_context=...,
    output_path=...,
    product_type=WorldCerealProduct.CROPTYPE,
    croptype_parameters=croptype_parameters,
)
```

I also changed the order of CLI parameters for the ´cropland_mapping.py` function, as I think it makes more sense to first define spatial/temporal context and after the product type and output file. `--epsg` is sent to the end (however it can be placed in any position by the user) as it is common practice.

### DIFF
--- a/src/worldcereal/job.py
+++ b/src/worldcereal/job.py
@@ -5,15 +5,12 @@ from pathlib import Path
 from typing import Optional, Union
 
 import openeo
-from openeo import DataCube
 from openeo_gfmap import BackendContext, BoundingBoxExtent, TemporalContext
-from openeo_gfmap.features.feature_extractor import apply_feature_extractor
-from openeo_gfmap.inference.model_inference import apply_model_inference
-from openeo_gfmap.preprocessing.scaling import compress_uint8, compress_uint16
 from pydantic import BaseModel
 
 from worldcereal.openeo.feature_extractor import PrestoFeatureExtractor
 from worldcereal.openeo.inference import CroplandClassifier, CroptypeClassifier
+from worldcereal.openeo.mapping import _cropland_map, _croptype_map
 from worldcereal.openeo.preprocessing import worldcereal_preprocessed_inputs_gfmap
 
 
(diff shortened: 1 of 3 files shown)


src/worldcereal/job.py
```diff
@@ -24,44 +24,108 @@ class WorldCerealProduct(Enum):
...
+
+class ClassifierParameters(BaseModel):
+    """Parameters for the classifier. Types are enforced by Pydantic.
+
+    Attributes
+    ----------
+    classifier_url : str
+        Public URL to the classifier model. Te file should be an ONNX accepting
+        a `features` field for input data and returning two output probability
+        arrays `true` and `false`.
```
kvantricht, line 53: in case of multiclass, it's not just `true` and `false`, no?

src/worldcereal/job.py
```diff
@@ -88,6 +152,8 @@ def generate_map(
     backend_context: BackendContext,
     output_path: Optional[Union[Path, str]],
     product_type: WorldCerealProduct = WorldCerealProduct.CROPLAND,
+    cropland_parameters: CropLandParameters = CropLandParameters(),
+    croptype_parameters: CropTypeParameters = CropTypeParameters(),
```
kvantricht, line 156: it's not confusing to a user that even if `product_type=='cropland'`, there is `croptype_parameters` given?

GriffinBabe: Indeed, I thought about it but I didn't find a trivial solution for this.

Maybe one solution would be to change the type from
`croptype_parameters: CropTypeParameters = CropTypeParameters()`
to 
`croptype_parameters: Optional[CropTypeParameters] = CropTypeParameters()`

In terms of logic this would change nothing practically but at least it is more clear for the user. I don't want to set it to `None` by default because that would throw off users that try to run croptype mapping without digging in the parameter's code (I assume most users want to use the default parameters when starting using our package).

I can also document this aspect in the docstring of the `generate_map` function

kvantricht: OK let's go for that for now

kvantricht (approved):

PR #106, opened 2024-07-08
Custom tile size

VincentVerelst:
@kvantricht , there was already an option in GFMap to set a custom tile_size when loading in collections, so I've just added an optional argument in `worldcereal_preprocessed_inputs_gfmap` to set a tilesize


kvantricht (commented):
Isn't it safer to explicitly state `if tile_size is not None` instead of doing a boolean evaluation?

VincentVerelst:
> Isn't it safer to explicitly state `if tile_size is not None` instead of doing a boolean evaluation?

aha yes, thanks for the suggestion! changes have been made

kvantricht (approved):
Great, thanks!

PR #113, opened 2024-08-08
79 s1 extractions

GriffinBabe:
More changes on the extraction scripts to come, but I'm merging this branch as it's been open for too long

- Code improvement by deleting duplicate code
- Improving the extraction scripts by setting up a notification system for smartphone 
- Various debugs, especially with s1 catalogue querries


scripts/extractions/extract_common.py
```diff
@@ -0,0 +1,191 @@
...
+
+        ref_id = extracted_gpd.iloc[idx].ref_id
+        valid_time = extracted_gpd.iloc[idx].valid_time
+        h3_l3_cell = extracted_gpd.iloc[idx].h3_l3_cell
+        s2_tile = row.s2_tile
+
+        item_asset_path = Path(list(item.assets.values())[0].href)
+
+        # Add some metadata to the result_df netcdf file
+        new_attributes = {
```
kvantricht, line 70: @GriffinBabe not sure if I fully understand this, but the `extract_common.py` seems to hold some common functionality like post job action, but the default parameters below are all related to S1. What's the reason for this? If some script does not provide a required parameter, the S1 information will be used? Won't that lead to confusion?

scripts/extractions/extract_common.py
```diff
@@ -0,0 +1,191 @@
...
+        orbit_state = ""
+
+    s2_tile_id = row.s2_tile
+    h3_l3_cell = row.h3_l3_cell
+    epsg = s2_grid[s2_grid.tile == s2_tile_id].iloc[0].epsg
+
+    subfolder = root_folder / ref_id / h3_l3_cell / sample_id
+    return (
+        subfolder
+        / f"{row.out_prefix}{orbit_state}_{sample_id}_{epsg}_{row.start_date}_{row.end_date}{row.out_extension}"
```
kvantricht, line 117: What is `orbit_state` for non-S1 extractions?

kvantricht (commented):
Just two minor questions on defaults in the `extract_common.py` script.

scripts/extractions/extract_common.py
GriffinBabe: Yes indeed, makes sense I'm changing this

scripts/extractions/extract_common.py
GriffinBabe: by default the value is an empty string  `""` so it doesn't impact the file name if there is no orbit information in the `row: pd.Series` variable

kvantricht (approved):
Thanks!

PR #115, opened 2024-08-12
extraction fixes - again

GriffinBabe:
* Added extract value option and orbit management in post-job action.
* Added connection errors as a catched exception


GriffinBabe:
Closes #119

GriffinBabe:
This was merged into branch "121-add-slope"

PR #123, opened 2024-08-20
fixed typo in the s1 item_asset definitions in extract_sar.py

VincentVerelst:
@GriffinBabe, just a small typo fix.

However, it's important to note that this typo makes it so that `load_stac` doesn't work. Luckily it's at collection level, so it can be easily adjusted for the extractions that are done already.


GriffinBabe (approved):

GriffinBabe:
Thanks @VincentVerelst !

Issue #119, opened 2024-08-13
Extracted patches have weird nodata value


GriffinBabe:
Extracted patches from the optical pipeline are written in uint16 data value but have a nodata value at 32769

Check if this can't be fixed in post_job_action

GriffinBabe:
PR #128

kvantricht:
What was the reason for this?

Issue #55, opened 2024-06-12
Duplicated band name remapping procedures in `worldcereal-classification` and in `presto-worldcereal`


kvantricht:
This is in fact a bit an overhead, if later on in `presto-worldcereal` we again remap: https://github.com/WorldCereal/presto-worldcereal/blob/main/presto/inference.py#L41

So we should consider to immediately use the presto naming here and remove the remapping in `presto-worldcereal`

_Originally posted by @kvantricht in https://github.com/WorldCereal/worldcereal-classification/pull/48#discussion_r1634829180_

PR #128, opened 2024-08-23
121 add slope & Improvement of S1 extractions

GriffinBabe:
This PR includes advancements on two features. It should have been decomposed in two PRs actually, but I was working on both features at the exact same time.

## Improvement of the S1 extractions

- Now using the sampleID to match the job output item to the reference DataFrame, instead of using the order on which the polygons and job items are provided in the inputs/outputs.
- Changed ´executor-memoryOverhead´ to new parameter ´python-memory´
- Post-job action
- Other small changes such as not computing the job DataFrame in ´extract.py´ if there is already the `job_tracking.csv` file present.

## Altitude and Slope parameters

- DEM parameter changed to "altitude" to be picked up by presto.
- Added a slope calculation, adding the `richdem` dependency.
- Updated dependencies ´worldcereal_deps.zip`

@kvantricht Before approving/merging, could you update the `worldcereal_deps.zip` file in the cloudferro bucket?

`https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/dependencies/worldcereal_deps.zip` => `https://s3.waw3-1.cloudferro.com/swift/v1/project_dependencies/worldcereal_deps.zip`


src/worldcereal/openeo/feature_extractor.py
```diff
@@ -38,7 +39,7 @@ class PrestoFeatureExtractor(PatchFeatureExtractor):
         "S2-L2A-B12": "B12",
         "S1-SIGMA0-VH": "VH",
         "S1-SIGMA0-VV": "VV",
-        "COP-DEM": "DEM",
+        "COP-DEM": "elevation",
```
kvantricht, line 41: In fact, we assume `DEM-alt-20m` which is later remapped to `elevation`: https://github.com/WorldCereal/presto-worldcereal/blob/main/presto/inference.py#L66

However, this may not be consistent with the other gfmap naming schemes. So it's fine by me to keep it like this for now, but I'll need to make a change in the presto code for that.

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -62,6 +63,25 @@ def output_labels(self) -> list:
         of the presto embeddings"""
         return [f"presto_ft_{i}" for i in range(128)]
 
+    def _compute_slope(self, inarr: xr.DataArray) -> xr.DataArray:
+        """Computes the slope using the richdem library. The input array should
+        have the following bands: 'elevation' And no time dimension. Returns a
+        new DataArray containing the new `slope` band.
+        """
+        from richdem import (  # pylint: disable=import-outside-toplevel
```
kvantricht, line 70: is this handled in the dependencies somewhere?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -62,6 +63,25 @@ def output_labels(self) -> list:
...
+            TerrainAttribute,
+            rdarray,
+        )
+
+        dem = inarr.sel(bands="elevation").values
+        dem_array = rdarray(dem, no_data=65535)
```
kvantricht, line 76: have you tested if the `no_data` value is handled correctly? Should we add a test for this to be sure?

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -110,6 +130,12 @@ def execute(self, inarr: xr.DataArray) -> xr.DataArray:
             get_presto_features,
         )
 
+        slope = self._compute_slope(inarr.isel(t=0))
```
kvantricht, line 132: I'm wondering why you chose to compute slope in the Presto feature extractor. I guess there's no other way to do it, like in where we actually assemble inputs? That's maybe not a UDF? It just feels a bit weird that slope computation is part of presto feature extractor.

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -62,6 +63,25 @@ def output_labels(self) -> list:
...
+            rdarray,
+        )
+
+        dem = inarr.sel(bands="elevation").values
+        dem_array = rdarray(dem, no_data=65535)
+        slope = TerrainAttribute(dem_array, attrib="slope_riserun")
+        return xr.DataArray(
+            slope[None, :, :],
+            dims=("bands", "x", "y"),
+            coords={"bands": ["slope"], "x": inarr.x, "y": inarr.y},
```
kvantricht, line 81: also here the name is not (yet) compatible with presto inference but this can be changed in presto code.

src/worldcereal/openeo/feature_extractor.py
kvantricht: Also, this means each time we run presto feature extractor (by default twice for cropland vs croptype, but even many more times in developing the algorithms locally), slope has to be recomputed. Would be nicer to have slope as part of the preprocessed inputs. But so I guess it won't be possible in an efficient way?

src/worldcereal/openeo/feature_extractor.py
GriffinBabe: How do you prefer really? I doesn't change for me if we name it here or in the presto library

src/worldcereal/openeo/feature_extractor.py
GriffinBabe: There's no process that does it. We could create a brand new UDF just for this aspect but it would be a heavier task. Also  not sure that it's very efficient. Also when you think about it other features are computed within this UDF from the Presto library such as lat/lon

src/worldcereal/openeo/feature_extractor.py
kvantricht: Ideally we'd immediately address this one: https://github.com/WorldCereal/worldcereal-classification/issues/55

kvantricht: So let's go for presto names `altitude` and `slope` like you did. Will do the other band name remapping of #55 in separate PR, linked to a change in presto code too.

src/worldcereal/openeo/feature_extractor.py
GriffinBabe: I remember that you asked me to verify this but I ended up dismissing it because unlike in the satio wrapper, here you can specify the `no_data` parameter.

You did good to ask because I checked and indeed it does accepts 65535 as nodata, but the output still uses -9999...

I added a fix as well as a small  test for this.

kvantricht: OK nice!

kvantricht (approved):
For me it looks good now. I will approve though we need to wait for dependency update. You can merge when that's done (and then we test).

GriffinBabe:
Closes #119 upon merging

PR #133, opened 2024-08-26
Presto-compilation

kvantricht:
Allow to parse the `compile` flag to Presto, which enables presto compilation of the encoder.


scripts/inference/collect_inputs.py
GriffinBabe, line 1: Can we rename `collect_inputs.py`, `cropland_mapping_local.py` and `croptype_mapping_local.py` by adding the `test_`  prefix in the names? All those "scripts" are actually more tests than utilities. We could even consider moving them in the test folder, what do you think?

src/worldcereal/job.py
```diff
@@ -339,5 +344,11 @@ def collect_inputs(
     inputs.execute_batch(
         outputfile=output_path,
         out_format="NetCDF",
-        job_options={"driver-memory": "4g", "executor-memoryOverhead": "4g"},
+        job_options={
```
GriffinBabe, line 347: Isn't the presto compile feature only available for more modern python versions than 3.8? If yes how do we specify which UDF environments to use?

kvantricht: It's not the python version itself that is the issue on openEO, it's the fact that their 3.8 image does not contain a compiler. I'm not sure if we need to take care of that here? Seems more an openEO thing. In the backlog is a task to create dependencies for 3.11 so we can test it on openEO.

scripts/inference/collect_inputs.py
kvantricht: for the local one I agree. The other two (regardless of their names which indeed could be improved) are not necessarily tests. I use them to actively run inference or collect inputs. The latter I'm doing now also with the updated extraction script, but if it's just for a single bbox, then these are still the go-to. You have another opinion?

GriffinBabe (approved):

PR #137, opened 2024-08-29
Add WorldCereal input extractions worfklow

kvantricht:


scripts/extractions/patch_extractions/extract.py
```diff
@@ -321,6 +363,12 @@ def manager_main_loop(
         default=1,
         help="The value of the `extract` flag to use in the dataframe.",
     )
+    parser.add_argument(
+        "--disable_stack",
```
GriffinBabe, line 367: I think it should be `disable_stac`

scripts/extractions/patch_extractions/extract_worldcereal.py
```diff
@@ -0,0 +1,398 @@
...
+    return cube.create_job(
+        out_format="NetCDF",
+        title=f"GFMAP_Extraction_WORLDCEREAL_{s2_tile}_{valid_time}",
+        sample_by_feature=True,
+        job_options=job_options,
+        feature_id_property="sample_id",
+    )
+
+
+def get_worldcereal_product(worldcereal_product, bounds, epsg):
```
GriffinBabe, line 213: Missing type hints here

scripts/extractions/patch_extractions/extract_worldcereal.py
```diff
@@ -0,0 +1,398 @@
...
+    vrt_file = f"/vitodata/worldcereal_data/MAP-v3/2021/VRTs/ESA_WORLDCEREAL_{worldcereal_product}_V1.vrt"
+
+    data = load_reproject(vrt_file, bounds, epsg)
+    data[np.isnan(data)] = 255
+    data = data.astype(np.uint8)
+
+    return data
+
+
+def postprocess_extracted_file(item_asset_path, new_attributes):
```
GriffinBabe, line 240: Same for here

scripts/extractions/patch_extractions/extract_worldcereal.py
```diff
@@ -0,0 +1,398 @@
...
+
+        # Postprocess file
+        postprocess_extracted_file(item_asset_path, new_attributes)
+
+    return job_items
+
+
+def generate_output_path_worldcereal(
+    root_folder: Path, geometry_index: int, row: pd.Series, s2_grid: gpd.GeoDataFrame
+):
```
GriffinBabe, line 385: To return the output type hint

GriffinBabe (changes requested):
Just some minor remarks. We could setup mypy in our linting actions which makes static checks on the type hints, but it will require some refactoring to make it compliant and I'm not sure it's really a huge issue for now

kvantricht:
> Just some minor remarks. We could setup mypy in our linting actions which makes static checks on the type hints, but it will require some refactoring to make it compliant and I'm not sure it's really a huge issue for now

shouldn't be hard to setup (can copy from presto, where we do it already), but as you say, it's gonna raise some issue the first time ;-) Maybe I can setup a branch for this specifically. It's gonna be worth it at some point.

scripts/extractions/patch_extractions/extract_worldcereal.py
kvantricht: @GriffinBabe returns nothing because eventually replaces the original file. Type hint should be -> None then?

GriffinBabe (approved):

PR #138, opened 2024-08-29
Compute slope with `scipy`, avoiding the use of `richdem`

kvantricht:


kvantricht:
@GriffinBabe this version now no longer needs `richdem`. You'll see `_compute_slope` is now much longer so not sure if it still fits in `FeatureExtractor`. Please have a look. Also, once global slope is available as collection, we need to implement a check that only computes slope if the band is not available yet (should be straightforward, can do this later in new branch).

One comment in the code now: correct slope computation needs to know the resolution. Set it at 10m now. Is it safe to assume this will always be the case? Or can we check it from the array?

GriffinBabe (approved):

PR #155, opened 2024-09-11
139-slope-collection

kvantricht:
Load slope from global collection when on CDSE backend


pyproject.toml
```diff
@@ -41,7 +41,7 @@ dependencies = [
     "numpy<2.0.0",
     "netcdf4<=1.6.4",
     "h5netcdf>=1.1.0",
-    "openeo>=0.29.0",
+    "openeo>=0.31.0",
```
GriffinBabe, line 44: I'll wait for this ticket to close before changing the OpenEO version

https://github.com/Open-EO/openeo-gfmap/issues/151

src/worldcereal/openeo/feature_extractor.py
```diff
@@ -144,6 +146,41 @@ def _rolling_fill(darr, max_iter=2):
 
             return _rolling_fill(darr, max_iter=max_iter)
 
+        def _downsample(arr, factor):
+            """
+            Downsamples a 2D NumPy array by a given factor with average resampling and reflect padding.
+
+            Parameters:
```
GriffinBabe, line 153: Let's keep the same docstyle by adding the "-------" under `Parameters`

GriffinBabe: Same for `Returns:`

src/worldcereal/openeo/preprocessing.py
```diff
@@ -244,7 +244,40 @@ def raw_datacube_DEM(
...
+    Returns
+    -------
+    DataCube
+        openEO datacube with the DEM data (and slope if available).
+    """
+
+    cube = extractor.get_cube(connection, spatial_extent, None)
+    cube = cube.rename_labels(dimension="bands", target=["elevation"])
+
+    if backend_context.backend.name == "CDSE" and fetch_type == FetchType.TILE:
```
GriffinBabe, line 260: better to do the comparison directly with the backend enum

`if backend_context.backend == Backend.CDSE`

GriffinBabe (commented):
Some minor remarks, except for the OpenEO client version

Also I would move the slope computation directly to the PatchFeatureExtractor superclass in GFMAP in the future, as it can be useful to everyone

pyproject.toml
kvantricht: ok but we can't merge this PR without bumping the version because the new slope collection cannot be loaded with `0.29.0`. Something in the python client changed and the former version cannot work with it.

src/worldcereal/openeo/feature_extractor.py
kvantricht: argh didn't use the automatic docstring formatting. Will fix this.

pyproject.toml
GriffinBabe: Okay, I'll try to close this issue GFMAP today

GriffinBabe: @kvantricht Nevermind sorry! this issue will come with Version 0.32 and the feature is not yet merged on the client repository 

- https://github.com/Open-EO/openeo-python-client/pull/607

kvantricht:
> Some minor remarks, except for the OpenEO client version
> 
> Also I would move the slope computation directly to the PatchFeatureExtractor superclass in GFMAP in the future, as it can be useful to everyone

yeah but the way it's implemented now (especially with the required change to force computation at 20m) is probably not generic enough. I think what would be better instead is a slope (or terrain attributes) process in openEO, which can then be accessed in GFMAP in a transparent way.

GriffinBabe (approved):

PR #157, opened 2024-09-13
Adding postprocessing to WorldCereal jobs

kvantricht:


GriffinBabe:
@kvantricht Made some good progress on this issue today 

A few things to discuss:

* We cannot go through the GFMAP connection methods as CDSE firewall is blocking our long UDFs. For the moment I commented that part and set to https://openeo.creo.vito.be/openeo/, which does backdoors the firewall. We need to find a solution for that before merging this PR of course
* There is one part missing for the postprocessing: whenever performing reclassification, we get the index from `np.argmax(...)`. However, we do not have access to a LUT as in the catboost inference. One way could be to read this information from the catboost model outside the UDFs in the `generate_map` function and  then provide the LUT to all UDFs from the parameters, I propose we discuss this tomorrow.
* I uploaded new unit-tests which are passing well. Could you provide me the preprocessed inputs NetCDF that was set-up here  https://github.com/WorldCereal/worldcereal-classification/blob/98-add-postprocessing/tests/worldcerealtests/conftest.py#L16 ? They are not tracked in the repository, I tried with my own files, with which I had a good crop-land but a bad crop-type (only other  crops predicted). Before trying on a different location I would like to test it with the files that you used here.

kvantricht:
> * Before trying on a different location I would like to test it with the files that you used here.

- this URL used to be the one we were using, I forgot why. This was indeed the reason. Let's see how it can be solved ...
- TBD tomorrow, not fully sure what the situation is
- Added the resource for cropland. For crop type I didn't have one. Should be simple to generate one though. TBD.

kvantricht:
@GriffinBabe in fact I don't think these are very useful right now. Bands are still only `['classification', 'probability']`. We need updated ones with all probabilities right?

kvantricht:
is `save_result()` now always triggered when requesting postprocessing? I can imagine e.g. for the demo that users just want the end product. The only use case for saving intermediate results is probably the global processing where we still want to play with the probabilities afterwards.

GriffinBabe:
Let's make it parametrizable in `generate_map` with default=False?

GriffinBabe:
@kvantricht 
* I added an exception in the `generate_map` function to connect to the backdoor URL if the users wants to connect to CDSE
* For the output labels of the model, and the index to class code LUT I decided to load those informations in a separate util function and pass those values in the UDFs parameters. Anyways before we needed to load the model locally in `CropTypeClassifier` to determine the number of output bands to output and specify in the computing graph, now I do it outside the UDF, but I also pass it to the `PostProcessor` so it can reclassify with the correct labels instead of the argmax indices.

src/worldcereal/job.py
```diff
@@ -209,6 +72,8 @@ def generate_map(
         Parameters for the croptype product inference pipeline. Only required
         whenever `product_type` is set to `WorldCerealProduct.CROPTYPE`, will be
         ignored otherwise.
+    postprocess_parameters: Optional[PostprocessParameters]
```
kvantricht, line 75: outdated docstring for this parameter?

src/worldcereal/job.py
```diff
@@ -238,7 +103,12 @@ def generate_map(
         raise ValueError(f"Format {format} not supported.")
 
     # Make a connection to the OpenEO backend
-    connection = BACKEND_CONNECTIONS[backend_context.backend]()
+    if backend_context.backend == Backend.CDSE:
+        connection = openeo.connect(
+            "https://openeo.creo.vito.be/openeo/"
```
kvantricht, line 108: should we add a comment here why we go this way for the time being or not needed?

src/worldcereal/openeo/inference.py
```diff
@@ -28,7 +28,7 @@ def dependencies(self) -> list:
         return []  # Disable the dependencies from PIP install
 
     def output_labels(self) -> list:
-        return ["classification", "probability"]
+        return ["classification", "max_probability"]
```
kvantricht, line 31: when individual probabilities are not included, should we call this just `probability` for clarity? Or does this give issues downstream?
edit: see next comment, I'm confused.

src/worldcereal/openeo/inference.py
```diff
@@ -42,18 +42,18 @@ def predict(self, features: np.ndarray) -> np.ndarray:
...
-        prediction_values = prediction_values * 100.0
-        prediction_values = np.round(prediction_values).astype("uint8")
-
-        return np.stack([binary_labels, prediction_values], axis=0)
+        # Extract all probabilities
+        all_probabilities = np.round(
+            np.array([[x["False"], x["True"]] for x in outputs[1]]) * 100.0
+        ).astype(np.uint8)
+        max_probability = np.max(all_probabilities, axis=1, keepdims=True)
+        # return np.concatenate(
```
kvantricht, line 53: i don't really understand. This was temporary version of me. We're now not returning all probabilities, so how can we smooth downstream?

src/worldcereal/openeo/inference.py
```diff
@@ -120,22 +130,26 @@ def predict(self, features: np.ndarray) -> np.ndarray:
...
         outputs = self.onnx_session.run(None, {"features": features})
 
-        # Get info on classes from the model
-        class_params = eval(
-            self.onnx_session.get_modelmeta().custom_metadata_map["class_params"]
-        )
-
         # Get classes LUT
-        LUT = dict(zip(class_params["class_names"], class_params["class_to_label"]))
+        lookup_table = self._parameters["lookup_table"]
```
kvantricht, line 134: what's the behavior if this key is not in the parameters?

src/worldcereal/openeo/inference.py
```diff
@@ -120,22 +130,26 @@ def predict(self, features: np.ndarray) -> np.ndarray:
...
 
-        return np.stack([labels, probabilities], axis=0)
+        # Extract per class probabilities
+        output_probabilities = []
+        for output_px in outputs[1]:
+            output_probabilities.append(list(output_px.values()))
+
+        output_probabilities = (np.array(output_probabilities) * 100).astype(np.uint8)
+
+        return np.hstack(
```
kvantricht, line 150: if no postprocessing is requested, we'll still get all the probabilities back? Shouldn't we then just get the default [classification, probability]?

src/worldcereal/openeo/mapping.py
```diff
@@ -124,7 +152,69 @@ def _croptype_map(
     if cropland_mask is not None:
         classes = classes.mask(cropland_mask == 0, replacement=254)
 
+    # Postprocess
+    if postprocess_parameters.enable:
+        if postprocess_parameters.save_intermediate:
+            classes = classes.save_result(format="GTiff")
+        # TODO: check how 254 value is treated during postprocessing
```
kvantricht, line 159: will this be checked before we merge?

src/worldcereal/openeo/mapping.py
```diff
@@ -124,7 +152,69 @@ def _croptype_map(
...
+
     # Cast to uint16
     classes = compress_uint16(classes)
 
     return classes
+
+
+def _postprocess(
+    classes: DataCube,
+    postprocess_parameters: "PostprocessParameters",
```
kvantricht, line 172: string still required? Thought you mentioned cyclical imports were no longer a problem?

src/worldcereal/openeo/postprocess.py
```diff
@@ -0,0 +1,153 @@
+import xarray as xr
+from openeo_gfmap.inference.model_inference import ModelInference
+
+
+class PostProcessor(ModelInference):
+    """Perform post-processing from the model outputs. Expects an input cube
+    with 2 + N bands, where N is the number of classes. The first band is the
```
kvantricht, line 7: is this requirement checked somewhere based on number of classes in LUT?

src/worldcereal/openeo/postprocess.py
```diff
@@ -0,0 +1,153 @@
...
+        or a multi-class classification model (croptype). Default is False.
+    lookup_table: Optional[dict]
+        Required if is_binary is False. A lookup table to map the class names
+        to class labels, ordered by model output.
+    """
+
+    EXCLUDED_VALUES = [254, 255, 65535]
+
+    def output_labels(self) -> list:
+        return ["classification", "max_probability"]
```
kvantricht, line 23: see earlier comment, we should probably call the final result just "probability". Besides, we may want to build in the option to still save all probabilities to the result in case we want crop probability maps.

src/worldcereal/openeo/postprocess.py
```diff
@@ -0,0 +1,153 @@
...
+
+        excluded_mask = np.in1d(
+            base_labels_vals.reshape(-1),
+            cls.EXCLUDED_VALUES,
+        ).reshape(*base_labels_vals.shape)
+
+        new_labels_vals = np.argmax(probabilities.values, axis=0)
+        new_max_probs_vals = np.max(probabilities.values, axis=0)
+
+        # TODO we are now reclassifying as integers, but we should use the LUT
```
kvantricht, line 86: when is this TODO tackled?

src/worldcereal/openeo/preprocessing.py
```diff
@@ -269,10 +269,11 @@ def raw_datacube_DEM(
                 "https://stac.openeo.vito.be/collections/COPERNICUS30_DEM_SLOPE",
                 spatial_extent=spatial_extent,
                 bands=["Slope"],
-            )
-            .rename_labels(dimension="bands", target=["slope"])
-            .min_time()
+            ).rename_labels(dimension="bands", target=["slope"])
+            # .min_time()
```
kvantricht, line 273: I'm actually not sure this should still be commented. It was a workaround for a metadata issue. I think it should be back in.

src/worldcereal/openeo/preprocessing.py
```diff
@@ -269,10 +269,11 @@ def raw_datacube_DEM(
...
         )
+        slope.metadata = slope.metadata.add_dimension("t", "2020-01-01", "temporal")
```
kvantricht, line 275: and in that case, this should no longer be required.

tests/worldcerealtests/test_inference.py
```diff
@@ -35,12 +36,12 @@ def test_cropland_inference(WorldCerealPreprocessedInputs):
 
     assert list(cropland_classification.bands.values) == [
         "classification",
-        "probability",
+        "max_probability",
```
kvantricht, line 39: cfr earlier comment. Should maybe rename.

src/worldcereal/job.py
GriffinBabe: I wouldn't publicly comment "We do this to avoid the firewall" haha

src/worldcereal/openeo/inference.py
GriffinBabe: Here we are in cropland classification, which is binary classification so you only need one probability to know the other. There is a parameter for the postprocessor called "is_binary" that takes this into account.

src/worldcereal/openeo/inference.py
GriffinBabe: I added a check at the beginning of the UDF

src/worldcereal/openeo/postprocess.py
GriffinBabe: Removed, It's done with `lookup_table`

src/worldcereal/openeo/preprocessing.py
GriffinBabe: There is an issue with CDSE that doesn't come with FED

With FED there is a temporal dimension detected and min_time drops it. In CDSE there is also a temporal dimension, but the openeo client is somehow unsynchronized and thinks there isn't (while the backend has it), so this is a workaround to allow it

GriffinBabe: I updated that part, maybe if you read it again it will make more sense

src/worldcereal/openeo/postprocess.py
GriffinBabe: I can add that option, do you want to keep the old probabilities (before postprocessing) or the ones after?

src/worldcereal/job.py
kvantricht: With this comment, you just did 😆

src/worldcereal/openeo/inference.py
kvantricht: ah ok that's once way. Not sure if that's easier to understand than just implementing the same functionality in a binary classifier instead of dealing with it in a postprocessor. But I guess ok for now. Should we remove the commented code here then? Some of the code I wrote is now actually convoluted if individual probabilities are not used.

src/worldcereal/openeo/inference.py
kvantricht: @GriffinBabe  what about this one?

src/worldcereal/openeo/postprocess.py
kvantricht: The ones after I guess, otherwise we can use the intermediate saving on the classification, not?

GriffinBabe: This was added

src/worldcereal/openeo/preprocessing.py
GriffinBabe: @kvantricht which version of the OpenEO python client do you have? I have `0.29.0`

GriffinBabe (approved):

Issue #125, opened 2024-08-21
Setup ground truth geoparquet generation pipeline using RDM API


kvantricht:
Create a function and an openEO job:

## The function:
Creates a GeoParquet of points samples and matching metadata:
* Either from a local dataset of polygons/points (WorldCereal Harmonized from @cbutsko)
* Either from a spatial/temporal extent used to query the RDM

The function implements a sampling of points from polygons with a strategy defined by the R&D team (start with centroid). And returns a GeoParquet of points with minimal metadata (GT code, SampleID, ...)

## OpenEO job:
Takes as input a GeoParquet of points to sample with additional metadata (GT, sampleID...)

From the collections of Sentinel1 and Sentinel2 patches, create a job that from input points samples:
* Sentinel 1 and Sentinel 2 data from the local stac collections
* METEO data from the pre-processed STAC collection available on CDSE bucket
* DEM data from global collection available on CDSE

Then performs processing such as monthly composites. Finally, concatenate GT and other metadata in a post-job action, returning a final GeoParquet with all the sample information alongside monthly-composited signal data and METEO/DEM.

kvantricht:
> Additional note: the _GeoParquet of points with minimal metadata (GT code, SampleID, ...)_ should be uploaded to somewhere like Artifactory, so that it can be read with openeo.

based on what? Extractions that are finished?

VincentVerelst:
After discussions with the openeo devs, we have decided to use the `load_url` process to load the reference data in openeo. For this two features need to be implemented on the VITO/CDSE backend:

- https://github.com/Open-EO/openeo-geopyspark-driver/issues/860
- https://github.com/Open-EO/openeo-geopyspark-driver/issues/859

GriffinBabe:
Also depends on
- https://github.com/WorldCereal/worldcereal-classification/issues/151

I'm taking care of this this week

HansVRP:
The decision has been made to develop this on the python client side. 
It is very specific to world cereal, so it will end up in the world cereal repo. 

Underlying technology to evaluate is duckdb/pandas.

requirement: location to store the geoparquet. (need to be accessible by terrascope back-end)
--> DIscuss with Peter/Jeroen

What defines success: Speed and efficiency

jdries:
@VincentVerelst  can you close or provide status update?

VincentVerelst:
@jdries , there's an open PR for this issue. Kristof will review after the WC demo.

PR #161, opened 2024-09-19
First version of a rdm api interaction module #125

VincentVerelst:
This is a first version of how we can construct a GeoParquet file from the RDM API, based on a user-defined AOI. 

How it works:

```python
import shapely
from worldcereal.rdm_api import query_rdm

poly = shapely.Polygon([(1.824467, 50.34406), (1.824467, 50.950124), (3.735661, 50.950124), (3.735661, 50.34406), (1.824467, 50.34406)])

gdf = query_rdm(poly)
```

A visualization of the output:

image

All feedback is welcome!


src/worldcereal/utils/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""This sub-module contains utilitary function and tools for worldcereal-classification"""
+
+from worldcereal.utils.rdm_interaction import query_ground_truth
+
+__all__ = [
+    "query_ground_truth",
```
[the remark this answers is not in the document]
VincentVerelst: One suggestion: we could make an entirely separate module for RDM API interaction. In `utils` have a `query_public_refdata` and `query_private_refdata` and let the latter interact with the RDM API interaction module?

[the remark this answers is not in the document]
VincentVerelst: Ok! Suggestions for the naming of the module always welcome :) (`rdm_interaction`, `rdm_api_interaction`, `rdm_api`, ...)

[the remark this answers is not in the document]
kvantricht: Shorter feels better, `rdm_api` might be good?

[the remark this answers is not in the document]
VincentVerelst: Ok, updated the folder structure

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -0,0 +1,143 @@
...
+    """
+
+    combined_query = ""
+    columns_str = ", ".join(columns)
+
+    for i, url in enumerate(urls):
+        query = f"""
+            SELECT {columns_str}, ST_AsWKB(geometry) AS wkb_geometry
+            FROM read_parquet('{url}')
+            WHERE ST_Within(geometry, ST_GeomFromText('{str(poly)}'))
```
kvantricht, line 99: as discussed, we have to check how this behaves with partial intersections.

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -0,0 +1,143 @@
...
+from pathlib import Path
+from typing import List, Union
+
+import duckdb
+import geopandas as gpd
+import requests
+from shapely import Polygon, wkb
+
+# Define the default columns to be extracted from the RDM API
+DEFAULT_COLUMNS = [
```
kvantricht, line 12: We propose these as the defaults: `['sample_id', 'ewoc_code', 'valid_time', 'quality_score_lc', 'quality_score_ct']`

kvantricht (commented):
Looks really great! TBD how we can cope with authenticated calls to RDM which is one more addition that we probably need at this point.

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -0,0 +1,180 @@
...
+        """
+        if i == 0:
+            combined_query = query
+        else:
+            combined_query += f" UNION ALL {query}"
+
+    return combined_query
+
+
+def query_ground_truth(
```
kvantricht, line 138: Maybe the only thing left is a better name for the main method. `public_query_rdm`? Or if we think we'll expand later on with options for authenticated queries, maybe just `query_rdm`?

kvantricht (approved):

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -0,0 +1,180 @@
...
+    geometry: BaseGeometry,
```
kvantricht, line 139: we're not enforcing any check on CRS. But I guess we have to be providing geometries in WGS84 projection, right? We should at least document it. I guess we cannot really check it if a base geometry doesn't even include something like a CRS.

src/worldcereal/rdm_api/rdm_interaction.py
kvantricht, line 138: I think it's more interesting to have this method return the `GeoDataFrame` so it can also be used without having to write the file to disk immediately. Another method can still do so if needed, but in my tests it's more useful to not write it to disk but keep in memory.

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -0,0 +1,180 @@
...
+    """Queries the RDM API and generates a GeoParquet file of all intersecting sample IDs.
+
+    Parameters
+    ----------
+    geometry : BaseGeometry
+        A user-defined polygon.
+    output_path : Union[str, Path]
+        The output path for the GeoParquet file.
+    temporal_extent : List[str], optional
+        A list of two strings representing the temporal extent, by default None. If None, all available data will be queried.
```
kvantricht, line 153: please document the expected format for the strings.

src/worldcereal/rdm_api/rdm_interaction.py
VincentVerelst: Currently, right before the writing to parquet, there is also the 'sampling' part, which in this case is just taking the centroid:
```python
gdf.geometry = (
    gdf.geometry.centroid
)  # Ensure that the geometry is a point, can be replaced by a more advanced algorithm
```

Maybe we can also leave this out, so that a user can apply a more advanced sampling method if desired? 

We then just return a GeoDataFrame where all geometries are fully within the user-specified geometry. (Geometries which were only partially in have been cut-off).

kvantricht: ah yes this is a good point I even overlooked. Yes because in my case I need the original geometries indeed.

VincentVerelst:
@kvantricht , quite some changes since your last review:

With the new changes made around authentication, a way to interact with the RDM API would now be as follows: 

```python
from worldcereal.rdm_api import RdmInteraction

interaction = RdmInteraction().authenticate()

gdf = interaction.query_rdm(geometry=multi_polygon, temporal_extent=temporal_extent)
```
Authentication now works similary to the OpenEO Python Client. You can also leave out `authenticate()`, in which case the rdm query will only interact with public datasets. 

For a STAC collection of patch extractions, you can extract the geometries of all patchtes into one `MultiPolygon` as follows (the following code will probably become part of the patch-to-points extraction script):

```python
import pystac
from shapely.geometry import shape, MultiPolygon

collection = pystac.read_file('/path/to/stac/collection.json')

polygons = []

for item in collection.get_items():  
    polygons.append(shape(item.geometry).buffer(1e-9))  # Add buffer to avoid TopologyException

multi_polygon = MultiPolygon(polygons)
temporal_extent = [collection.extent.temporal.intervals[0][0], collection.extent.temporal.intervals[0][1]]
```

The results for a query on this MultiPolygon and temporal_extent look like this (cached 64x64 patches in pink, RDM query results in purple):
image

pyproject.toml
kvantricht, line 1: can you pull in current `main`? There's a merge conflict but I think it's just the order of dependencies.

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -0,0 +1,250 @@
...
+        Returns
+        -------
+        List[str]
+            A List containing the HTTPs URLs of the GeoParquet files for each collection ID.
+        """
+        urls = []
+
+        for id in collection_ids:
+            url = f"{self.RDM_ENDPOINT}/collections/{id}/download"
+            response = requests.get(url, headers=self._get_headers())
```
kvantricht, line 145: do we need some resilience here? what's the timeout?

kvantricht (approved):
I just added two comments, other than that this looks good to merge so we can move on to the next step. Nice job!

VincentVerelst:
@kvantricht, added some resilience and a timeout to the RDM interaction. That should cover your final comments!

kvantricht:
Perfect, thanks!

PR #162, opened 2024-09-19
Made the extractor main's loop robust to artifactory errors

GriffinBabe:


kvantricht (commented):
Could you tell me what exactly has now changed with this PR?

GriffinBabe:
> Could you tell me what exactly has now changed with this PR?

Miscellaneous stuff to make the extractor more robust:
* Catch exceptions coming from the Artifactory and restart pipeline with that (previously this only was done with OpenEO errors)
* Check in the post-job action to avoid crash whenever there are more that one row with the sample sampleID (this shouldn't happen, but rarely there are some rows in the input dataframe that are duplicate)
* One error in the `pystac.Collection` object that I don't understand myself, but adding the attribute "strategy" fixes it

kvantricht (approved):

PR #164, opened 2024-09-20
Discussion thread for patch to point notebook

VincentVerelst:


notebooks/patch_to_point/patch_to_point.ipynb
```diff
@@ -0,0 +1,689 @@
...
+   "execution_count": 1,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create a list of the paths to the split collections\n",
+    "import glob\n",
+    "s1_collection_paths = glob.glob(\"/data/users/Public/vincent.verelst/world_cereal/s1_s2_meteo_extraction/Sentinel1/split_stac/*/collection.json\")\n",
+    "s2_collection_paths = glob.glob(\"/data/users/Public/vincent.verelst/world_cereal/s1_s2_meteo_extraction/Sentinel2/small-split-stac/*/collection.json\")\n",
+    "\n",
+    "assert len(s1_collection_paths) == len(s2_collection_paths), \"Mismatch in number of collections\""
```
kvantricht, line 61: we can imagine that often these collections are not fully in sync. One may still be extracting, so this assertion could fail. Wouldn't a better option be to somehow make an intersect of the two collections and only keep those that are "identical" for both inputs?

notebooks/patch_to_point/patch_to_point.ipynb
```diff
@@ -0,0 +1,689 @@
...
+   "metadata": {},
+   "source": [
+    "### Create_datacube"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Next, create the `create_datacube` function to be given to the GFMapJobManager is created. For this we cannot use worldcereal_classification's `worldcereal_preprocessed_inputs` in it's current form. Maybe best to discuss how we unify as much as possible the preprocessing of the different scripts.\n",
```
kvantricht, line 212: This is indeed a key point! We need to keep these as aligned as much as possible.

notebooks/patch_to_point/patch_to_point.ipynb
```diff
@@ -0,0 +1,689 @@
...
+    "\n",
+    "METEO is loaded from a pre-composited STAC collection --> we have to think about how we keep this up-to-date (maybe include a trigger in the automatic RDM API interaction?)\n",
```
kvantricht, line 214: what do you mean by this? and can you remind me why we couldn't go for the normal daily meteo collection instead of the precomposited one?

notebooks/patch_to_point/patch_to_point.ipynb
```diff
@@ -0,0 +1,689 @@
...
+    "            \"S2-L2A-B03\",\n",
+    "            \"S2-L2A-B04\",\n",
+    "            \"S2-L2A-B05\",\n",
+    "            \"S2-L2A-B06\",\n",
+    "            \"S2-L2A-B07\",\n",
+    "            \"S2-L2A-B08\",\n",
+    "            \"S2-L2A-B8A\",\n",
+    "            \"S2-L2A-B11\",\n",
+    "            \"S2-L2A-B12\"\n",
+    "        ]  # Not extracting SCL and distance to cloud, can easily be added if needed\n",
```
kvantricht, line 259: so how are we applying the mask_scl_dilation cloud mask then now?

notebooks/patch_to_point/patch_to_point.ipynb
```diff
@@ -0,0 +1,689 @@
...
+    "    meteo_cube.result_node().update_arguments(featureflags={\"tilesize\": 1})\n",
+    "    meteo_cube = meteo_cube.rename_labels(\n",
+    "            dimension=\"bands\", target=[\"AGERA5-PRECIP\", \"AGERA5-TMEAN\"]\n",
+    "        )\n",
+    "\n",
+    "    cube = s1_cube.merge_cubes(s2_cube)\n",
+    "    cube = cube.merge_cubes(dem_cube)\n",
+    "    cube = cube.merge_cubes(meteo_cube)\n",
+    "    \n",
+    "    # This is just a placeholder for the RDM API interaction function. In this example we just load in some random points from parquet as an example.\n",
```
kvantricht, line 287: these random points should be the sampling procedure in the ground truth I guess? What will happen if a point somehow falls outside the available patches?

notebooks/patch_to_point/patch_to_point.ipynb
VincentVerelst: This would make it difficult to track which 'part' of the STAC collection has been already extracted to points and which not. Would it not be safer to wait until all S1 and S2 patch extractions are finished for a certain dataset before we trigger the point extractions?

notebooks/patch_to_point/patch_to_point.ipynb
VincentVerelst: Can we add this to the agenda to discuss in our meeting on monday?

notebooks/patch_to_point/patch_to_point.ipynb
VincentVerelst: Currently the meteo data is saved as a static STAC collection, which is not automatically updated. In the future, if reference data becomes available for 2025, there will be no meteo data available in this STAC collection for that temporal extent. We therefore have to make sure it stays up to date. 

The reason we use a pre-composited STAC collection, is because once we merge meteo with S1 and S2, it will be upsampled to a 10/20m resolution. If we then still need to apply the compositing, it will be a huge increase in credit cost and computing time.

notebooks/patch_to_point/patch_to_point.ipynb
VincentVerelst: Is the mask not already applied in the patch extractions? https://github.com/WorldCereal/worldcereal-classification/blob/main/src/worldcereal/openeo/preprocessing.py#L369

notebooks/patch_to_point/patch_to_point.ipynb
VincentVerelst: The patches are constructed as: reference geometry (from public RDM) -> take centroid -> buffer to square of 64x64
The points are constructed as: same reference geometry (from public RDM) -> take centroid
So by construction the points wshould always fall within the patches. 

For custom user reference data, the points may fall outside of the patches, but it was my understanding that we would not use the cached patch data for these extractions.

notebooks/patch_to_point/patch_to_point.ipynb
kvantricht: do they all come with their own collections? I think i don't get the entire picture on collection organization yet and how the growing database of extractions will be handled. Yes we should wait till the extractions for "a dataset" are done, but aren't there other ways as well to check if points are already extracted from patches? Looking into the existing geoparquet e.g.? let's brainstorm a bit about this on monday.

notebooks/patch_to_point/patch_to_point.ipynb
kvantricht: yes!

notebooks/patch_to_point/patch_to_point.ipynb
kvantricht: ok I see, indeed in this case we need to automate the meteo compositing (it seems extracting meteo in the end could have been also a solution worth investigating, because it now turns out to be a painful thing too? ;-))

notebooks/patch_to_point/patch_to_point.ipynb
kvantricht: @GriffinBabe is that so? I though the point of extracting the scl dilation mask and add it to extractions was that we could still switch to e.g. the DL-based mask once it would become mature. Does it now turn out we're masking already with a fixed mask?!

notebooks/patch_to_point/patch_to_point.ipynb
kvantricht: ah yeah ok but we'll add another point sampling method afterwards to not always have to stick (exclusively) to centroids. We just need to make sure we then don't make mistakes and fall outside the patches.

kvantricht: This sampling will be an important thing, to be aligned with @cbutsko how we can do this correctly.

notebooks/patch_to_point/patch_to_point.ipynb
VincentVerelst: Ah sorry, my bad! For the patch extractions the masking is turned off: https://github.com/WorldCereal/worldcereal-classification/blob/main/scripts/extractions/patch_extractions/extract_optical.py#L127

I'll add the scl mask as an extra band then.

kvantricht: Just in time to save me from some bad sleep 🤣

notebooks/patch_to_point/patch_to_point.ipynb
kvantricht: From discussion with @cbutsko we should find a way to flag the original center polygon('s centroid) as the initially targeted sample, and then add the "collateral" fields('s centroids) from the patch with another flag.

kvantricht:
@VincentVerelst I think we can close this PR and remove this branch as everything is already merged regarding patch to point?

PR #170, opened 2024-10-01
Refactoring of System V1 demo

kvantricht:


src/worldcereal/utils/upload.py
```diff
@@ -0,0 +1,212 @@
...
+    @classmethod
+    def from_openeo_connection(cls, conn: Connection) -> AWSSTSCredentials:
+        """
+        Takes an OpenEO connection object and returns temporary credentials to interact with S3
+        """
+        auth_token = conn.auth.bearer.split("/")
+        os.environ["AWS_ENDPOINT_URL_STS"] = cls.STS_ENDPOINT
+        sts = boto3.client("sts")
+        return AWSSTSCredentials._from_assume_role_response(
+            sts.assume_role_with_web_identity(
```
pvbouwel, line 85: @kvantricht  Same limits currently apply as https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/sts/client/assume_role_with_web_identity.html

So default duration, which is used at the moment,  is 1 hour. you can bump it to 12 hours by passing `DurationSeconds=43200` here.  If 12 hours is not sufficient please provide a duration that would be acceptable.

kvantricht: @jdries you were hinting at a bit more than 12 hours but maybe for now it is sufficient?

src/worldcereal/utils/upload.py
```diff
@@ -0,0 +1,213 @@
...
+        config = TransferConfig(multipart_threshold=self.MULTIPART_THRESHOLD_IN_MB * MB)
+
+        bucket = self.BUCKET_NAME
+        key = self.get_upload_key(object_name)
+
+        self.get_s3_client().upload_file(src_file_path, bucket, key, Config=config)
+        return f"s3://{bucket}/{key}"
+
+    def get_presigned_url(
+        self, s3_uri: str, expires_in_seconds: int = 3600 * 24
```
pvbouwel, line 153: You could increase the lifetime of the presigned URL to 6 days rather than 1 day.

src/worldcereal/utils/upload.py
pvbouwel: After discussing with @jdries I understood the important part is the presigned URLs. They actually have a longer lifetime and you could increase them to 6 days (so on https://github.com/WorldCereal/worldcereal-classification/pull/170/files#r1784775538 )

src/worldcereal/utils/upload.py
kvantricht: Done in [c0b2140](https://github.com/WorldCereal/worldcereal-classification/pull/170/commits/c0b2140be4aedf526b8b9cc70af78b2cfc4ee132)

src/worldcereal/utils/upload.py
kvantricht: Done in [c0b2140](https://github.com/WorldCereal/worldcereal-classification/pull/170/commits/c0b2140be4aedf526b8b9cc70af78b2cfc4ee132)

PR #175, opened 2024-10-04
allow multiple outputs for inference results

jdegerickx:
When saving an intermediate result in openeo, one needs to specify the "filename_prefix" option in the "save_result" function.
This was not done before, so the asset was overwritten by the final result.

I tested this for cropland inference and it works now, successfully generating 2 assets.

Instead of downloading the assets during the "execute_batch" call (which only supports download of a single asset), I have introduced a few lines which download all assets belonging to a certain job, including the metadata json file.

I changed the nature of InferenceResults, to be able to cope with multiple outputs for a single inference run.
To do so, I renamed WorldCerealProduct to WorldCerealProductType (because this was actually a product type) and created a new data class WorldCerealProduct which now holds the URL, local path and product type of an actual product (raster file).

All changes were tested for the cropland inference case.
I was not yet able to test inference for crop type, as the job failed due to another reason...


src/worldcereal/job.py
```diff
@@ -167,21 +185,52 @@ def generate_map(
...
-    asset = job.get_results().get_assets()[0]
+    # Get job results
+    job_result = job.get_results()
+
+    # Get the products
+    assets = job_result.get_assets()
+    products = {}
+    for asset in assets:
+        name = asset.name.split("_")[0]
+        prod_type = (
```
kvantricht, line 209: no need to abbreviate variable name

src/worldcereal/job.py
```diff
@@ -146,7 +164,7 @@ def generate_map(
             .reduce_dimension(
                 dimension="t", reducer="mean"
             )  # Temporary fix to make this work as mask
-        )
+        ).save_result(format="GTiff", options=dict(filename_prefix="cropland-mask"))
```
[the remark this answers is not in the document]
kvantricht: regarding the naming, see comment a bit below. Product should probably be called the same way. Cropland could be fine for now.

src/worldcereal/job.py
```diff
@@ -167,21 +185,52 @@ def generate_map(
...
+    job_result = job.get_results()
+
+    # Get the products
+    assets = job_result.get_assets()
+    products = {}
+    for asset in assets:
+        name = asset.name.split("_")[0]
+        prod_type = (
+            WorldCerealProductType.CROPLAND
+            if name.split("-")[0] == "cropland"
```
kvantricht, line 211: this looks a bit odd an not so future proof (when more products would be added, this thing is gonna get ugly). Maybe we should in the filename use the exact match of the name of the product, such as `CROPLAND` or `CROPTYPE`. This way, you can directly construct the `WorldCerealProductType` from the name and avoid `if-else` statements.

src/worldcereal/job.py
```diff
@@ -167,21 +185,52 @@ def generate_map(
...
+            else WorldCerealProductType.CROPTYPE
+        )
+        if output_path is not None:
```
kvantricht, line 214: When is this None?

src/worldcereal/openeo/mapping.py
```diff
@@ -76,7 +76,9 @@ def _cropland_map(
     # Postprocess
     if postprocess_parameters.enable:
         if postprocess_parameters.save_intermediate:
-            classes = classes.save_result(format="GTiff")
+            classes = classes.save_result(
+                format="GTiff", options=dict(filename_prefix="cropland-raw")
```
kvantricht, line 80: also here we need to be careful about hardcoding specific naming schemes. We should take this from the `WorldCerealProductType` otherwise the naming convention loses consistency and gets scattered around.

src/worldcereal/openeo/mapping.py
```diff
@@ -155,7 +157,9 @@ def _croptype_map(
     # Postprocess
     if postprocess_parameters.enable:
         if postprocess_parameters.save_intermediate:
-            classes = classes.save_result(format="GTiff")
+            classes = classes.save_result(
+                format="GTiff", options=dict(filename_prefix="croptype-raw")
```
kvantricht, line 161: Same comment as above on naming.

kvantricht (commented):
Nice evolution! But I added quite a few comments to think about and we can discuss on Monday. File naming consistency is something we have to think about carefully.

src/worldcereal/job.py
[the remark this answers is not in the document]
jdegerickx: Only when the user requests intermediate products, the cropland mask will be saved now.
I made sure the name matches and is inferred from WorldCerealProductType. (so only options are cropland-raw and cropland-cleaned.

cropland or tempoarycrops --> we can still decide on that. I just kept what was implemented by Darius...

Yes, if you request postprocessing AND saving of intermediate results, you will now get a total of 4 products in the crop type workflow...

src/worldcereal/job.py
jdegerickx: fixed in new commit

src/worldcereal/job.py
jdegerickx: fixed in commit. The WorldCerealProductType is now constructed from the asset name without the if - else statement

src/worldcereal/job.py
jdegerickx: output_path (now output_dir) is an optional argument in the generate_map function.
If it is left unspecified, it will default to None and no products will be downloaded locally.

src/worldcereal/openeo/mapping.py
jdegerickx: hardcoding of product names avoided in new commit. 
WorldCerealProductType is not an argument in the _cropland_map function, so I did need to refer to WorldCerealProductType.CROPLAND to define the name

src/worldcereal/openeo/mapping.py
jdegerickx: same comment as above

src/worldcereal/openeo/mapping.py
```diff
@@ -76,7 +77,10 @@ def _cropland_map(
     # Postprocess
     if postprocess_parameters.enable:
         if postprocess_parameters.save_intermediate:
-            classes = classes.save_result(format="GTiff")
+            product_type = WorldCerealProductType.CROPLAND.value
+            classes = classes.save_result(
+                format="GTiff", options=dict(filename_prefix=f"{product_type}-raw")
```
kvantricht, line 82: same comment, the product_type can probably directly be inserted here.

src/worldcereal/openeo/mapping.py
```diff
@@ -155,7 +159,10 @@ def _croptype_map(
     # Postprocess
     if postprocess_parameters.enable:
         if postprocess_parameters.save_intermediate:
-            classes = classes.save_result(format="GTiff")
+            product_type = WorldCerealProductType.CROPTYPE.value
+            classes = classes.save_result(
+                format="GTiff", options=dict(filename_prefix=f"{product_type}-raw")
```
kvantricht, line 164: same comment, the product_type can probably directly be inserted here.

src/worldcereal/openeo/mapping.py
```diff
@@ -155,7 +159,10 @@ def _croptype_map(
...
+            )
         # TODO: check how 254 value is treated during postprocessing
```
kvantricht, line 166: Tis TODO is there for a while. You've been looking into postprocessing already. Could you check for this comment and once verified remove the TODO?

src/worldcereal/openeo/mapping.py
kvantricht: Makes sense, within `_cropland_map` we can safely assume that that is indeed the product type.

kvantricht (approved):
good to go!

PR #177, opened 2024-10-10
Postproc

jdegerickx:
- added majority voting as a potential postprocessing pipeline (default remains smooth_probabilities)
- bug fixing when downloading assets
- "-cleaned" now removed from filename downloaded asset
- various minor fixes and updates on date slider, get seasons and ui_map
- added product post processing after download (split into classification and confidence .tif with LUT and color maps)
- is_binary argument in crop type pipeline should depend on lookup_table?


notebooks/utils.py
```diff
@@ -379,3 +371,137 @@ def train_classifier(inputs, targets):
...
+
+NODATAVALUE = {
+    "cropland": 255,
+    "croptype": 255,
+    "confidence": 255,
+}
+
+
+COLORMAP = {
+    "active-cropland": {
```
kvantricht, line 392: Why do we need this here?

src/worldcereal/job.py
```diff
@@ -194,19 +201,28 @@ def generate_map(
...
-        filename_prefix=filename,
+        filename_prefix=product_type.value,
     )
 
+    # Get look-up tables
+    luts = {}
+    luts[WorldCerealProductType.CROPLAND.value] = load_model_lut(
+        cropland_parameters.classifier_parameters.classifier_url
+    )
+    if (
```
kvantricht, line 218: there's no way to avoid if statements here I guess?

src/worldcereal/openeo/postprocess.py
```diff
@@ -32,6 +33,139 @@ def output_labels(self) -> list:
...
 
+    @classmethod
+    def majority_vote(
+        cls,
+        base_labels: xr.DataArray,
+        max_probabilities: xr.DataArray,
+        kernel_size: int,
+        conf_threshold: int,
+    ) -> xr.DataArray:
+        """
```
kvantricht, line 44: can you reformat the docstring following the Numpy style format? Your autodocstring should be configured for that already I think?

src/worldcereal/openeo/postprocess.py
```diff
@@ -32,6 +33,139 @@ def output_labels(self) -> list:
     def dependencies(self) -> list:
         return []
 
+    @classmethod
+    def majority_vote(
```
kvantricht, line 37: This is tested for multiclass arrays?

src/worldcereal/openeo/postprocess.py
```diff
@@ -64,6 +198,12 @@ def smooth_probabilities(
             )
             probabilities_vals[class_idx][excluded_mask] = 0
 
+        # Sum of probabilities should be 1
+        probabilities_vals = probabilities_vals / probabilities_vals.sum(axis=0)
+
+        # Cast back to uint16
+        probabilities_vals = np.round(probabilities_vals * 100.0).astype("uint16")
```
kvantricht, line 205: can we do this in one go?
something like
np.round(`probabilities_vals / probabilities_vals.sum(axis=0) * 100.0).astype(np.uint16)`?

src/worldcereal/utils/map.py
```diff
@@ -1,51 +1,14 @@
...
-        3: (247, 185, 29, 255),  # millet_sorghum
-        4: (186, 186, 186, 0),  # other_crop
-        5: (167, 245, 66, 255),  # rapeseed
-        6: (85, 218, 218, 255),  # soy
-        7: (245, 66, 111, 255),  # sunflower
-        8: (186, 113, 53, 255),  # wheat
-    },
-    "probabilities": get_probability_cmap(),
-}
+LOCALTILESERVER_PORT = 8889
```
kvantricht, line 11: what do we still need this for? Does this local tile server stuff work? I thought we were not going to try to go this far.

notebooks/utils.py
jdegerickx: is a remainder from phase I, will delete...

src/worldcereal/openeo/postprocess.py
jdegerickx: yes, tested it for multiclass crop type...

src/worldcereal/utils/map.py
jdegerickx: moving product visualization to utils.py...

src/worldcereal/job.py
jdegerickx: not that I can think of no...

tests/worldcerealtests/test_postprocessing.py
kvantricht, line 1: you could maybe add two or three validation tests for the parameters.
Like [here](https://github.com/WorldCereal/worldcereal-classification/blob/6dfa75824b7772a3a10f6d30c11331bbef834f49/tests/worldcerealtests/test_preprocessing.py#L20), you can check if an error is raised as expected, and if correct parameters just work.

kvantricht (approved):
One more comment on adding a test

PR #183, opened 2024-10-14
Speedup extractions from public Parquet

cbutsko:


kvantricht (approved):
Reviewed and tested, works like a charm and fast, thanks!!

PR #184, opened 2024-10-14
Add default training pipeline

kvantricht:
This PR adds the training pipelines for the default cropland model trained on global Presto embeddings for the WorldCereal reference database.


scripts/spark/train_catboost.py
```diff
@@ -0,0 +1,528 @@
...
+
+        logger.info("Setting up model ...")
+
+        # Manually control class name order!
+        class_names = [
+            self.settings["classes"][class_nr]
+            for class_nr in range(len(self.settings["classes"]))
+        ]
+
+        model = CatBoostClassifier(
```
jdegerickx, line 334: should these model settings be configurable?

src/worldcereal/train/__init__.py
jdegerickx, line 1: all this filtering of reference datasets and filtering of samples based on rules does not feel generic.
It's also very much tuned to Phase I extractions.
We should probably get rid of this at some point?
Maybe even have a separate repository where we train the global models?

jdegerickx (changes requested):
main comment is whether we really want to commit all this Phase I related cleaning of datasets to this repository?
I guess in the end all default models will be trained based on Phase II RDM samples and extractions.
So perhaps training of global models should (for now) be done in a separate repository, where we import functionality from worldcereal-classification?
Just a suggestion...

src/worldcereal/train/__init__.py
kvantricht: At some point, yes maybe, but this how the model is currently trained so we need to be transparent. I agree that these methods need to evolve in the coming month.

kvantricht:
Closing as no longer relevant.

PR #185, opened 2024-10-14
Fix filenaming

jdegerickx:
- temporal dimension is stripped from classification result right after prediction UDF
- postprocessing no longer needs temporal dimension as input
- to have correct filenames, I needed temporal_extent as additional input to _cropland_map and _croptype_map functions
- "period" is now added as an attribute to each WorldCerealProduct

All changes were tested both for cropland and crop type and everything looked fine (both names of resulting files + visualization)


src/worldcereal/openeo/mapping.py
```diff
@@ -21,6 +22,7 @@
 
 def _cropland_map(
     inputs: DataCube,
+    temporal_extent: TemporalContext,
```
kvantricht, line 25: The docstring should be updated to cover this additional required argument

src/worldcereal/job.py
kvantricht, line 1: To be honest, I don't really understand the need for this `period` attribute. I would instead use the original `TemporalContext` parameter as attribute `temporal_extent` for the product. It is exactly meant for this purpose without the need to define an additional string-based attribute.

src/worldcereal/job.py
```diff
@@ -135,6 +139,10 @@ def generate_map(
         # Make a connection to the OpenEO backend
         connection = BACKEND_CONNECTIONS[backend_context.backend]()
 
+    # Get start and end dates for product file naming
+    start = temporal_extent.start_date.replace("-", "")
```
kvantricht, line 143: Why do the dashes have to go? Is that more readable? These two lines - if we go for the use of `temporal_extent` - are not needed. Can be plugged directly into `filename_prefix` constructor.

src/worldcereal/job.py
```diff
@@ -164,27 +173,29 @@ def generate_map(
...
+            postprocess_parameters=postprocess_parameters,
+        )
 
         # Save final mask if required
         if croptype_parameters.save_mask:
             cropland_mask = cropland_mask.save_result(
                 format="GTiff",
                 options=dict(
-                    filename_prefix=f"{WorldCerealProductType.CROPLAND.value}",
+                    filename_prefix=f"{WorldCerealProductType.CROPLAND.value}_{start}_{end}",
```
kvantricht, line 188: As mentioned above, can be directly taken from `temporal_extent`

src/worldcereal/job.py
```diff
@@ -203,12 +214,13 @@ def generate_map(
         JOB_OPTIONS.update(job_options)
 
     # Execute the job
+    filename_output = f"{product_type.value}_{start}_{end}"
```
kvantricht, line 217: Is a separate line needed for this? A bit inconsistent with the `cropland_mask` filename construction.

src/worldcereal/job.py
```diff
@@ -241,6 +253,7 @@ def generate_map(
         products[asset_name] = {
             "url": asset.href,
             "type": asset_type,
+            "period": f"{start}_{end}",
```
kvantricht, line 256: I suggest to change this to: `"temporal_extent": temporal_extent`

src/worldcereal/openeo/mapping.py
```diff
@@ -80,13 +82,18 @@ def _cropland_map(
         ],
     )
 
+    # Get rid of temporal dimension
+    classes = classes.reduce_dimension(dimension="t", reducer="mean")
+
     # Postprocess
     if postprocess_parameters.enable:
         if postprocess_parameters.save_intermediate:
+            start = temporal_extent.start_date.replace("-", "")
```
kvantricht, line 91: see earlier comment on dashes and the need yes/no to put this in separate line.

src/worldcereal/openeo/mapping.py
```diff
@@ -160,17 +170,22 @@ def _croptype_map(
...
+    classes = classes.reduce_dimension(dimension="t", reducer="mean")
+
     # Mask cropland
     if cropland_mask is not None:
         classes = classes.mask(cropland_mask == 0, replacement=254)
 
     # Postprocess
     if postprocess_parameters.enable:
         if postprocess_parameters.save_intermediate:
+            start = temporal_extent.start_date.replace("-", "")
```
kvantricht, line 183: Idem

kvantricht (changes requested):
My most important remark is on the introduction of `period` which I think is not needed if we stick to the already defined `TemporalContext`

jdegerickx:
Issues addressed. 
Tested for cropland, everything looks fine!

kvantricht (approved):
Looks good!

PR #188, opened 2024-10-15
get rid of conf_threshold parameter as it may introduce "undecided" p…

jdegerickx:
…ixels


src/worldcereal/openeo/postprocess.py
```diff
@@ -156,7 +150,7 @@ def majority_vote(
         # Setting excluded values back to their original values
         for excluded_value in cls.EXCLUDED_VALUES:
             aggregated_predictions[prediction == excluded_value] = excluded_value
-            aggregated_probabilities[prediction == excluded_value] = cls.NODATA
+            aggregated_probabilities[prediction == excluded_value] = excluded_value
```
kvantricht, line 153: maybe a stupid question, but does probabilities shape always match predictions shape? No matter the amount of classes?

jdegerickx: probablities here is only the max_probabilities, not the per-class probabilities

kvantricht (approved):

PR #189, opened 2024-10-15
Make the use of class weights optional (default=False)

kvantricht:


cbutsko (approved):
Looks good, didn't spot anything suspicious.

PR #190, opened 2024-10-15
cleaned notebooks for demo

jdegerickx:
- cleaned all outputs
- harmonized explanations across notebooks
- altered filenames
- fixed table of contents
- added figure on cropland sample density


notebooks/worldcereal_v1_demo_custom_cropland.ipynb
```diff
@@ -11,90 +11,56 @@
...
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# <a id='toc2_'></a>[Define a region of interest](#toc0_)\n",
+    "### 1. Define your region of interest\n",
     "\n",
     "When running the code snippet below, an interactive map will be visualized.\n",
     "Click the Rectangle button on the left hand side of the map to start drawing your region of interest.\n",
+    "\n",
     "Currently, there is a maximum size of 250 km² for your area within this demo. Upon exceeding this limit, an error will be shown.\n",
```
kvantricht, line 55: I think we should tackle it as follows:
- have a default max of 250km²
- if between 250km² and 750km² -> show the error but provide users with info how to manually raise the limit to max 750 km², pointing them to the credit cosrt
- if >750km², consider this as a hard maximum

notebooks/worldcereal_v1_demo_custom_cropland.ipynb
```diff
@@ -345,78 +198,40 @@
...
-   ],
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
    "source": [
     "from worldcereal.utils.upload import deploy_model\n",
     "from openeo_gfmap.backend import cdse_connection\n",
     "\n",
-    "model_url = deploy_model(cdse_connection(), custom_model, pattern=\"demo_cropland_TZA\")"
+    "modelname = <your_model_name>\n",
```
kvantricht, line 216: already wrap this in a string to make sure the user knows it's supposed to be a string?

kvantricht (commented):
Two comments put in one notebook but valid for the others too. Besides, the postprocessing parameter section seems overkill, especially if we have it everywhere like that.

notebooks/worldcereal_v1_demo_custom_cropland.ipynb
jdegerickx: used input method instead, asking user for a modelname (not containing spaces)
The modelname is then also used for naming the output directory

notebooks/worldcereal_v1_demo_custom_cropland.ipynb
jdegerickx: Currently, there is a maximum size of 250 km² for your area within this demo. Upon exceeding this limit, an error will be shown.
You can bypass this limit by altering the code below to:<br>
*map = ui_map(area_limit=750)*<br>

Processing areas beyond 750 km² are currently not supported.

kvantricht: don't forget to mention the actual reason for this: credit usage :-) otherwise a user may think it's odd why this limit is there in the first place

jdegerickx:
- added some more explanation on area limits for ui_map
- ensured user can set custom model name and folder name for each individual model run using new get_input() function

kvantricht (approved):

PR #195, opened 2024-10-15
194-align-training-window-with-user-defined-temporal-extent

cbutsko:


src/worldcereal/utils/refdata.py
```diff
@@ -132,12 +136,116 @@ def query_public_extractions(
...
 
     # Process the parquet into the format we need for training
-    processed_public_df = process_parquet(public_df_raw)
+    processed_public_df = process_parquet(public_df_raw, processing_period)
 
     return processed_public_df
 
 
-def process_parquet(public_df_raw: pd.DataFrame) -> pd.DataFrame:
+def month_diff(month1: int, month2: int) -> int:
```
kvantricht, line 144: I may be overlooking something (though I tested quite a bit), but can the same not be accomplished with a very simple function instead?

```
def month_diff(month1: int, month2: int) -> int:
    return month2 - month1 if month2 >= month1 else 12 - month1 + month2
```

Another comment: it's documented "the number of months between two given months". If I'd say the number of months between January and March, what do you expect? There's one month between them, February, the total span is three months, Jan-Feb-Mar. So think carefully about what the functionality you actually need, and then make sure it's documented like this as well.

src/worldcereal/utils/refdata.py
```diff
@@ -132,12 +136,116 @@ def query_public_extractions(
...
+    shift = [xx for xx in [shift1, shift2] if (xx > 0) and (xx <= 12)][0]
+    if shift == 12:
+        shift = 0
+
+    return shift
+
+
+def get_best_valid_date(row):
+    """
+    Determine the best valid date for a given row based on forward and backward shifts.
```
kvantricht, line 180: is this a standard Numpy docstring? Hard to see from github changes, just checking with you.

src/worldcereal/utils/refdata.py
```diff
@@ -154,7 +262,63 @@ def process_parquet(public_df_raw: pd.DataFrame) -> pd.DataFrame:
     from presto.utils import process_parquet as process_parquet_for_presto
 
     logger.info("Processing selected samples ...")
+
+    if processing_period is not None:
+        logger.info("Aligning the samples with the user-defined temporal extent ...")
+
+        # get the middle of the user-defined temporal extent
+        start_date, end_date = processing_period.to_datetime()
+        processing_period_middle_ts = start_date + pd.DateOffset(months=5)
```
kvantricht, line 271: the middle is hard to define with an even number of timesteps. Should we be aligned with e.g. [this](https://github.com/WorldCereal/worldcereal-classification/blob/1a394187a3595fd94ddbe1ac7fbc153d1ce79637/src/worldcereal/openeo/feature_extractor.py#L306)? And therefore rather add 6 months? I may be wrong here, was just wondering.

src/worldcereal/utils/refdata.py
```diff
@@ -154,7 +262,63 @@ def process_parquet(public_df_raw: pd.DataFrame) -> pd.DataFrame:
...
+        sample_dates = (
+            public_df_raw[["sample_id", "start_date", "end_date", "valid_date"]]
+            .drop_duplicates()
+            .reset_index(drop=True)
+        )
+
+        # save the true valid_date for later
+        true_valid_date_map = sample_dates.set_index("sample_id")["valid_date"]
+
+        # calculate the shifts and assignt new valid date
```
kvantricht, line 284: small typo in comment

kvantricht (changes requested):
Great work! Just few remarks for now.

PR #216, opened 2024-11-05
Consolidate extraction endpoints

### LINKED ISSUE #118: Consolidate extraction end points
I noticed that for point extractions we only split by S2 tile: https://github.com/WorldCereal/worldcereal-classification/blob/16e50a3fe19d9e6923417ccfc48af74e2e74f8b8/scripts/extractions/point_extractions/point_extractions.py#L241

While in the patch extractions we also split by year: https://github.com/WorldCereal/worldcereal-classification/blob/main/scripts/extractions/patch_extractions/extract.py#L98

Furthermore we're [computing ](https://github.com/WorldCereal/worldcereal-classification/blob/main/scripts/extractions/point_extractions/point_extractions.py#L88) the mean `valid_time` to determine extraction window. In case samples span multiple years this is unsafe.

It was already addressed in the `patch_extractions\extract.py`. After discussing with @GriffinBabe we think it's a good idea to step away from separate patch and point extraction folders and put the `point_extractions.py` alongside the other extract scripts in `patch_extractions`, reusing as much as possible joint functionality. 

As a final step we can get rid of `patch_extractions` folder alltogether and keep things clear and simple for the user.

VincentVerelst:
Closes #118


VincentVerelst:
@kvantricht , the changes discussed on Teams have been added by the way.

kvantricht (approved):
This looks good to merge! Could you add a follow-up issue to class-ify this in the future? Thanks!!

PR #218, opened 2024-11-07
tile size must now be multiple of 16 in openeo

jdegerickx:
In openeo there has been a change, implying that tile sizes need to be multiples of 16 now.
Inference runs started failing...
This small change fixes the issue.


kvantricht (approved):
Great, thank!!

PR #228, opened 2024-11-25
Write patch extraction metadata directly to STAC API

VincentVerelst:
- Recommended to add two new environment variables `STAC_API_USERNAME` and `STAC_API_PASSWORD`. If these don't exist, authentication will happen via authentication code flow, which will require manual intervention for each batch job, which is not very practical
- Currently URLs point to openeo-dev STAC API, untill we get the correct permissions for POST requests to the prod STAC API (WIP by DevOps)
- Two collections have been created: `worldcereal_sentinelx_patch_extractions`. We're in talks with the DevOps team to only give PUT, POST and DELETE permissions to the 'wcextractions' account, so that other users cannot accidentally overwrite the STAC metadata. To be seen if this is possible
- One blocking point on the prod STAC API is the non-working `/bulk_items/` endpoint. DevOps team is aware.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,8 +46,6 @@ dependencies = [
     "openeo-gfmap==0.2.0",  
     "pyarrow",  
     "pydantic==2.8.0",  
-    "pystac==1.10.1",
-    "pystac-client==0.8.3",
     "rioxarray>=0.13.0",  
     "scipy",
     "duckdb>=1.1.0",
@@ -71,7 +69,9 @@ train = [
   "scikit-learn==1.5.0",
   "torch==2.3.1",
   "ipywidgets==8.1.3",
-  "duckdb==1.1.0"
+  "duckdb==1.1.0",
+  "pystac==1.10.1",
+  "pystac-client==0.8.3"
 ]
 
 [tool.pytest.ini_options]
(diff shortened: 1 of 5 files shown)


src/worldcereal/openeo/extract.py
```diff
@@ -49,6 +54,8 @@ def post_job_action_patch(
     title: str,
     spatial_resolution: str,
     s1_orbit_fix: bool = False,  # To rename the samples from the S1 orbit
+    write_stac_api: bool = False,
+    sensor: str = "PATCH_SENTINEL1",
```
kvantricht, line 58: this value is different from the `"Sentinel1"` in `extract.py`. Is that intended?

kvantricht (commented):
Looks cool! Not easy for me to understand what happens just from code. Do we need tests for this functionality? Once this thing is kind of operational, we should sit together to have a strategy for ingesting previous extractions as well.

pyproject.toml
kvantricht, line 1: I'm wondering if these dependencies shouldn't rather be added to either the `train` or even the `dev` dependencies? Will someone external using the repo ever require this or is this pure internal stuff? In case of the former, they will only do so if they'll eventually train a model, in case of the latter it's even more only for dev I'd say. What do you think?

VincentVerelst: Makes sense to me, this is only used in the patch extractions, so will probably only be used by `wcextractions`

kvantricht (approved):
Very clean code!! I'll approve and let's see how it goes in practice.

PR #235, opened 2024-12-18
234-refactor-point-extractions

kvantricht:


scripts/extractions/extract.py
```diff
@@ -103,7 +103,7 @@ def prepare_job_dataframe(
     pipeline_log.info("Preparing the job dataframe.")
 
     # Filter the input dataframe to only keep the locations to extract
-    input_df = input_df[input_df["extract"] == extract_value].copy()
+    input_df = input_df[input_df["extract"] >= extract_value].copy()
```
VincentVerelst, line 106: Just for my information, was there a change in extract flag convention that prompted this change in the code?

scripts/extractions/point_extractions/extract_point_worldcereal.py
```diff
@@ -74,7 +90,6 @@ def create_job_dataframe_point_worldcereal(
             "start_date": start_date,
             "end_date": end_date,
             "s2_tile": s2_tile,
-            "h3_l3_cell": h3_l3_cell,
             "geometry": job.to_json(),
```
VincentVerelst, line 93: See my comment below. You can already add the `start_date` and `end_date` in the FeatureCollection that is created in this line to automatically have these columns in your final output parquet.

VincentVerelst (changes requested):

scripts/extractions/extract.py
kvantricht: it was in principle never supposed to be just binary. We aimed this to be something that includes (in the future) a priority depending on its value. To be on the safe side, this now extracts everything that is not 0.

scripts/extractions/point_extractions/extract_point_worldcereal.py
```diff
@@ -150,7 +173,24 @@ def post_job_action_point_worldcereal(
         gdf = gpd.read_parquet(item_asset_path)
 
         # Convert the dates to datetime format
-        gdf["date"] = pd.to_datetime(gdf["date"])
+        gdf["timestamp"] = pd.to_datetime(gdf["date"])
+        gdf.drop(columns=["date"], inplace=True)
+
+        # Derive latitude and longitude from the geometry
+        gdf["lat"] = gdf.geometry.y
+        gdf["lon"] = gdf.geometry.x
```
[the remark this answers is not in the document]
kvantricht: I don't understand this. A job can contain multiple points. `geometry` at that point is a `FeatureCollection`. How can I set there point-specific lat/lon attributes for the dataframe resulting from `aggregate_spatial`?

scripts/extractions/point_extractions/extract_point_worldcereal.py
kvantricht: as you can see, it is already set here too, because this is where we get temporal information for filtering collections from. So maybe I should test if indeed by doing this it's now preserved in the resulting dataframe, as before it was not the case.

VincentVerelst: I'm sorry for the confusion:

- Only properties of features in the FeatureCollection (`job_to_json()`) in the `geometry` column will be present in the final output geoparquet
- This is because this FeatureCollection is passed to `aggregate_spatial`
- In the code `job` is a `geopandas.GeoDataFrame`. So you first need to add `lat`, `lon`, `start_date` and `end_date` columns to `job` and then convert it to a FeatureCollection.

kvantricht: Now I got it. Is the new commit what you mean?

VincentVerelst: Did you push the commit already?

kvantricht: :facepalm: I did now

VincentVerelst: Aha I see it now, indeed that's what I meant :)

VincentVerelst (approved):

PR #237, opened 2024-12-18
introduce legend functionality + example script

jdegerickx:
Functionality to upload, download and delete .csv files to Artifactory containing the WorldCereal crop type legend.
Automatic versioning is applied and upon upload, the "latest" version is always updated on artifactory.

### DIFF
--- a/src/worldcereal/utils/legend.py
+++ b/src/worldcereal/utils/legend.py
@@ -1,0 +1,134 @@
+import json
+import os
+import subprocess
+from pathlib import Path
+
+from loguru import logger
+
+ARTIFACTORY_BASE_URL = (
+    "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/"
+)
+
+
+def _get_artifactory_credentials():
+    """Get credentials for upload and delete operations on Artifactory.
+    Returns
+    -------
+    tuple (str, str)
+        Tuple containing the Artifactory username and password.
+    Raises
+    ------
+    ValueError
+        if ARTIFACTORY_USERNAME or ARTIFACTORY_PASSWORD are not set as environment variables.
+    """
+
+    artifactory_username = os.getenv("ARTIFACTORY_USERNAME")
+    artifactory_password = os.getenv("ARTIFACTORY_PASSWORD")
+
+    if not artifactory_username or not artifactory_password:
+        raise ValueError(
+            "Artifactory credentials not found. "
+            "Please set ARTIFACTORY_USERNAME and ARTIFACTORY_PASSWORD environment variables."
+        )
+
+    return artifactory_username, artifactory_password
+
+
+def upload_legend_csv_artifactory(
+    srcpath: Path,
+    date: str,
+) -> str:
+    """Uploads a CSV file containing the worldcereal land cover/crop type legend to Artifactory.
+    Parameters
+    ----------
+    srcpath : Path
+        Path to csv file that needs to be uploaded to Artifactory.
+    date : str
+        Date tag to be added to the filename. Should be in format YYYYMMDD.
+    Returns
+    -------
+    str
+        artifactory download link
+    Raises
+    ------
+    FileNotFoundError
+        if srcpath does not exist
+    """
+    if not srcpath.is_file():
+        raise FileNotFoundError(f"Required file `{srcpath}` not found.")
+
+    # Get Artifactory credentials
+    artifactory_username, artifactory_password = _get_artifactory_credentials()
+
+    # We  upload the file with a specific date tag and also with a "latest" tag
+    target_names = [
+        f"WorldCereal_LC_CT_legend_{date}.csv",
+        "WorldCereal_LC_CT_legend_latest.csv",
+    ]
+    targetpaths = [f"{ARTIFACTORY_BASE_URL}legend/{n}" for n in target_names]
+
+    for targetpath in targetpaths:
+        logger.info(f"Uploading `{srcpath}` to `{targetpath}`")
+
+        cmd = (
+            f"curl -u{artifactory_username}:{artifactory_password} -T {srcpath} "
+            f'"{targetpath}"'
+        )
+
+        output, _ = subprocess.Popen(
+            cmd, stdout=subprocess.PIPE, shell=True
+        ).communicate()
+        decoded_output = output.decode("utf-8")
+
+        # Parse as JSON if applicable
+        try:
+            parsed_output = json.loads(decoded_output)
+        except json.JSONDecodeError as e:
+            raise ValueError(f"Failed to parse output as JSON: {decoded_output}") from e
+
+    # Access the desired value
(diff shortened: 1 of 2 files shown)


src/worldcereal/utils/legend.py
```diff
@@ -0,0 +1,134 @@
...
+        try:
+            parsed_output = json.loads(decoded_output)
+        except json.JSONDecodeError as e:
+            raise ValueError(f"Failed to parse output as JSON: {decoded_output}") from e
+
+    # Access the desired value
+    return parsed_output.get("downloadUri")
+
+
+def download_latest_legend_from_artifactory(download_path: Path) -> Path:
```
kvantricht, line 93: not sure yet how it's used, but it could be useful if we only want to access the contents, that you allow the option to download to temporary file that is automatically destroyed after the contents have been loaded. Because what do you want to happen ultimately? Do you want to get the contents of the CSV? THen why not immediately returning that instead of the path to the file?

kvantricht (changes requested):
This PR at the moment does not really interact with the legend file. Would be good to have functionality in that actually reads the contents of the CSV and also allows with a one-liner to immediately get the latest version from artifactory als in-memory legend with a tempfile being removed.

I'm also missing some retries, because artifactory can sometimes be buggy. Actually wondering if we should have a backup file added to the repo in case the retries fail as well? We should sometimes update this one too and definitely log a warning if this one is used because of issues with artifactory, but I'm just a bit concerned if we need to fully rely on artifactory for this without a backup.

src/worldcereal/utils/legend.py
```diff
@@ -34,6 +37,66 @@ def _get_artifactory_credentials():
     return artifactory_username, artifactory_password
 
 
+def _upload_with_retries(
```
kvantricht, line 40: suggest to rename to `_upload_file`

src/worldcereal/utils/legend.py
```diff
@@ -34,6 +37,66 @@ def _get_artifactory_credentials():
...
+    srcpath,
+    targetpath,
+    username,
+    password,
+    retries=3,
+    wait=2,
+) -> str:
+    """_summary_
```
kvantricht, line 48: looks like the summary still needs to be filled in

src/worldcereal/utils/legend.py
```diff
@@ -34,6 +37,66 @@ def _get_artifactory_credentials():
...
+            logger.error(f"Attempt {attempt + 1} failed: {e}")
+            if attempt < retries - 1:
+                time.sleep(wait)
+            else:
+                raise
+
+    raise RuntimeError("Failed to upload file")
+
+
 def upload_legend_csv_artifactory(
```
kvantricht, line 100: could it be just called `upload_legend` ?

src/worldcereal/utils/legend.py
```diff
@@ -68,34 +131,54 @@ def upload_legend_csv_artifactory(
...
-            parsed_output = json.loads(decoded_output)
-        except json.JSONDecodeError as e:
-            raise ValueError(f"Failed to parse output as JSON: {decoded_output}") from e
+    # Return the download link of latest uploaded file
+    return artifactory_link
 
-    # Access the desired value
-    return parsed_output.get("downloadUri")
 
+def get_latest_legend_from_artifactory() -> pd.DataFrame:
```
kvantricht, line 142: can we just call this `get_legend` ?

src/worldcereal/utils/legend.py
```diff
@@ -68,34 +131,54 @@ def upload_legend_csv_artifactory(
...
+
+    # clean up the legend file
+    legend = legend[legend["ewoc_code"].notna()]
+    drop_columns = [c for c in legend.columns if "Unnamed:" in c]
+    legend.drop(columns=drop_columns, inplace=True)
+
+    return legend
+
+
+def _download_latest_legend_from_artifactory(
```
kvantricht, line 167: can we just call this `_download_legend`?

src/worldcereal/utils/legend.py
```diff
@@ -68,34 +131,54 @@ def upload_legend_csv_artifactory(
...
+    download_path: Path,
```
kvantricht, line 168: suggest to change this to `dstpath` for consistency with your other functions.

src/worldcereal/utils/legend.py
```diff
@@ -105,30 +188,68 @@ def download_latest_legend_from_artifactory(download_path: Path) -> Path:
...
-
     download_path.mkdir(parents=True, exist_ok=True)
     download_file = download_path / latest_file
-
     cmd = f'curl -o {download_file} "{link}"'
 
-    subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True).communicate()
+    for attempt in range(retries):
+        try:
+            logger.info(
```
kvantricht, line 200: I can imagine this method being used in quite a few occasions. INFO level logging can in this case really clutter output logs. I would consider changing this to DEBUG level.

src/worldcereal/utils/legend.py
```diff
@@ -105,30 +188,68 @@ def download_latest_legend_from_artifactory(download_path: Path) -> Path:
...
+                f"Downloading latest legend file: {latest_file} (Attempt {attempt + 1})"
+            )
+            subprocess.Popen(cmd, stdout=subprocess.PIPE, shell=True).communicate()
+            logger.info("Download successful!")
```
kvantricht, line 204: Idem as previous comment.

src/worldcereal/utils/legend.py
```diff
@@ -105,30 +188,68 @@ def download_latest_legend_from_artifactory(download_path: Path) -> Path:
...
+
+            return download_file
 
-    return download_file
+        except subprocess.CalledProcessError as e:
+            logger.error(f"Attempt {attempt + 1} failed: {e}")
```
kvantricht, line 209: I would change to `logger.warning`

src/worldcereal/utils/legend.py
```diff
@@ -105,30 +188,68 @@ def download_latest_legend_from_artifactory(download_path: Path) -> Path:
...
+            if attempt < retries - 1:
+                time.sleep(wait)
+            else:
+                raise
```
kvantricht, line 213: and right before raising here doing `logger.error`, because this is when it will actually be treated as error.

kvantricht (changes requested):
I have a few comments for you to consider. One additional thing is the multiple repeated implementations of the retries. Would it be an option to implement this only once and reuse in your different methods? If that's too hard, fine to keep as is for now.

jdegerickx:
> I have a few comments for you to consider. One additional thing is the multiple repeated implementations of the retries. Would it be an option to implement this only once and reuse in your different methods? If that's too hard, fine to keep as is for now.

Thanks for the thorough review and nice suggestions!
I believe it all looks better now.

kvantricht (approved):

PR #238, opened 2024-12-18
demonstration on how to use cropland extent udp

jdegerickx:
I would suggest to add a script that documents the use of the crop extent UDP.
This script should be extended by @VincentVerelst, to also showcase how to override the default CRS of the output products.

### DIFF
--- a/scripts/inference/cropland_mapping_udp.py
+++ b/scripts/inference/cropland_mapping_udp.py
@@ -24,11 +24,15 @@ temporal_extent = ["2020-01-01", "2020-12-31"]
 
 c = openeo.connect("openeofed.dataspace.copernicus.eu").authenticate_oidc()
 
+# This argument is optional and determines the projection of the output products
+target_projection = "EPSG:4326"
+
 classes = c.datacube_from_process(
     process_id="worldcereal_crop_extent",
     namespace="https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/main/openeo_udp/worldcereal_crop_extent.json",
     temporal_extent=temporal_extent,
     spatial_extent=spatial_extent,
+    projection=target_projection,
 )
 
 ONNX_DEPS_URL = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/openeo/onnx_dependencies_1.16.3.zip"
--- a/scripts/inference/cropland_mapping_udp.py
+++ b/scripts/inference/cropland_mapping_udp.py
@@ -1,0 +1,49 @@
+"""This short script demonstrates how to run WorldCereal cropland extent inference through
+    an OpenEO User-Defined Process (UDP) on CDSE.
+
+    The WorldCereal default cropland model is used and default post-processing is applied
+    (using smooth_probabilities method).
+
+    The user needs to manually download the resulting map through the OpenEO Web UI:
+    https://openeofed.dataspace.copernicus.eu/
+    Or use the openeo API to fetch the resulting map.
+    """
+
+import openeo
+
+spatial_extent = {
+    "west": 622694.5968575787,
+    "south": 5672232.857114074,
+    "east": 623079.000934101,
+    "north": 5672519.995940826,
+    "crs": "EPSG:32631",
+    "srs": "EPSG:32631",
+}
+
+temporal_extent = ["2020-01-01", "2020-12-31"]
+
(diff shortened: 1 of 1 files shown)


kvantricht (commented):
UDPs will now be added to repo (separate PR). What is needed for this older one then?

jdegerickx:
I guess once UDP's are added to the repo in the other PR, this PR and branch can be deleted?

jdegerickx:
@VincentVerelst, would be nice if we could finalize this PR now that the UDP's are in our repo.
Could you check the cropland UDP script and make sure it is fully up-to-date?
Would be nice if we can show in this script what the parameters are that you as a user can play with. Not sure I captured them all at the moment

scripts/inference/cropland_mapping_udp.py
```diff
@@ -0,0 +1,53 @@
...
+temporal_extent = ["2020-01-01", "2020-12-31"]
+
+c = openeo.connect("openeofed.dataspace.copernicus.eu").authenticate_oidc()
+
+# This argument is optional and determines the projection of the output products
+target_projection = "EPSG:4326"
+
+classes = c.datacube_from_process(
+    process_id="worldcereal_crop_extent",
+    namespace="https://raw.githubusercontent.com/ESA-APEx/apex_algorithms/refs/heads/main/openeo_udp/worldcereal_crop_extent.json",
```
VincentVerelst, line 32: The UDPs have been removed from APEx and added to this repo. The latest version URL is: https://github.com/WorldCereal/worldcereal-classification/blob/worldcereal_crop_extent_v1.0.1/src/worldcereal/udp/worldcereal_crop_extent.json

For future reference: you can find the latest UDP by looking at the tags in this repository and opening the newest version (in this case: `worldereal_crop_extent_v1.0.1`). The older ones will also always keep existing by the way, so your code won't break each time we update the UDP.

scripts/inference/cropland_mapping_udp.py
```diff
@@ -0,0 +1,53 @@
...
+    "executor-memoryOverhead": "1g",
+    "python-memory": "3g",
+    "soft-errors": "true",
+    "udf-dependency-archives": [f"{ONNX_DEPS_URL}#onnx_deps"],
+}
+
+job = classes.execute_batch(
+    title="WorldCereal Crop Extent UDP test",
+    out_format="GTiff",
+    job_options=JOB_OPTIONS,
```
VincentVerelst, line 52: With the latest new updates in openEO, `job_options` don't have be set explicitly anymore. They are already embedded in the UDP. It's of course still possible, in case you want to overwrite some settings, like memory for example.

For demonstration purposes, it might be nice to leave some of them in (like `executor-memory`, `python-memory`), but mention that they are optional. Other job options like `udf-dependency-archives` are things users shouldn't have to worry about anymore.

VincentVerelst (changes requested):

jdegerickx:
@VincentVerelst, is there a particular reason why we need to go through openeo Federation backend?
Normal CDSE backend also works?

jdegerickx:
@VincentVerelst, this doesn't seem to work...
image

VincentVerelst:
@jdegerickx ,

- No, using the normal CDSE backend is indeed also fine. We're just pushing the openeofed backend as good practice, but for WorldCereal there should be no reason to just use CDSE 
- For the namespace you need to provide the raw github URL (this is the URL which only shows the raw content, not the full Github page: https://raw.githubusercontent.com/WorldCereal/worldcereal-classification/refs/tags/worldcereal_crop_extent_v1.0.1/src/worldcereal/udp/worldcereal_crop_extent.json

Let me know if that works!

jdegerickx:
> @jdegerickx ,
> 
> * No, using the normal CDSE backend is indeed also fine. We're just pushing the openeofed backend as good practice, but for WorldCereal there should be no reason to not just use CDSE
> * For the namespace you need to provide the raw github URL (this is the URL which only shows the raw content, not the full Github page: https://raw.githubusercontent.com/WorldCereal/worldcereal-classification/refs/tags/worldcereal_crop_extent_v1.0.1/src/worldcereal/udp/worldcereal_crop_extent.json
> 
> Let me know if that works!

Yes, works now, thanks!
@kvantricht, I tested the UDP and everything seems to work as expected.
Can you check and approve this PR?

kvantricht (approved):

VincentVerelst (approved):

PR #254, opened 2025-01-20
get_legend reads directly from URL + extended to IRR

jdegerickx:
Based on suggestion from @cbutsko, get_legend now reads the legend directly from the URL instead of first downloading it.
I kept the function download_legend as well just in case some applications would require downloading the actual file...

I also added support to download the irrigation legend. 
get_legend by default returns the crop type legend, but you can get the irrigation legend by specifying the "topic"

### DIFF
--- a/scripts/misc/legend.py
+++ b/scripts/misc/legend.py
@@ -4,7 +4,12 @@ Example script showing how to upload, download and delete the WorldCereal crop t
 
 from pathlib import Path
 
-from worldcereal.utils.legend import delete_legend_file, get_legend, upload_legend
+from worldcereal.utils.legend import (
+    delete_legend_file,
+    download_legend,
+    get_legend,
+    upload_legend,
+)
 
 if __name__ == "__main__":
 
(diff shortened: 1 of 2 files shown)


scripts/misc/legend.py
```diff
@@ -17,8 +22,17 @@
     # Upload the legend to Artifactory
     link = upload_legend(srcpath, date)
 
-    # Download the latest legend from Artifactory
+    # Get the latest legend from Artifactory (as pandas DataFrame)
     legend = get_legend()
 
+    legend_irr = get_legend(topic="irr")
```
kvantricht, line 28: Is it transparent enough towards a user to use this kind of abbreviations?

scripts/misc/legend.py
```diff
@@ -17,8 +22,17 @@
...
+
+    # Download the latest legend from Artifactory
+    legend_path = download_legend(Path("/vitodata/worldcereal/tmp/jeroen/"))
```
kvantricht, line 31: Avoid committing absolute paths to locations not reachable by typical users. By default I'd download to the current working directory (`"."`).

scripts/misc/legend.py
```diff
@@ -17,8 +22,17 @@
...
+
+    irr_legend_path = download_legend(
+        Path("/vitodata/worldcereal/tmp/jeroen/"), topic="irr"
```
kvantricht, line 34: Same comment, avoid using these paths in public code.

kvantricht (changes requested):
Looks good, except for the absolute paths to `vitodata` locations.

scripts/misc/legend.py
cbutsko: I agree. maybe using Literal type can be a good trade0ff here? smth like:
```
from typing import Literal
topic: Literal["crop", "irr"] = "crop"
```

kvantricht: yes, although I even mean to be more explicit: `irrigation` instead of `irr`. And I'm also not sure `crop` is fully in line with what it actually is, although this could still be fine.

jdegerickx: converted to Literal with options "landcover" and "irrigation"

kvantricht (approved):
Thanks!

Issue #233, opened 2024-12-17
Ensure RDM only returns samples with non-zero confidence scores


jdegerickx:
During quality checks, some confidence scores are set to zero.
Make sure we do not use those samples in the processing module

jdegerickx:
added functionality to filter on quality score when requesting samples from RDM.
Can be closed.

PR #260, opened 2025-01-22
#233 allow setting minimum quality score when requesting samples from…

jdegerickx:
When requesting samples from the RDM, a user is now able to set a minimum quality score for both land cover and crop type


kvantricht (approved):
Looks good!

PR #271, opened 2025-01-27
Create extract module

VincentVerelst:


kvantricht (commented):
Looks good! One question: as everything is now under `extract` module, could it make sense to rename all submodules, stripping the `extract_` part from them to avoid redundancy and long imports?

Eg. `src/worldcereal/extract/extract_common.py` > `src/worldcereal/extract/common.py`

VincentVerelst:
> Looks good! One question: as everything is now under `extract` module, could it make sense to rename all submodules, stripping the `extract_` part from them to avoid redundancy and long imports?
> 
> Eg. `src/worldcereal/extract/extract_common.py` > `src/worldcereal/extract/common.py`

Agreed, I've updated the filenames

PR #274, opened 2025-01-27
Rdm demo

jdegerickx:
Added new notebook to showcase interaction with RDM.

To increase user-friendliness, I took some tools from v1 notebooks and altered them slightly (ui_map, crop type picker). Especially the latter was built from scratch based on the worldcereal legend.

In the process I re-organized the notebook utils into separate files. V1 notebooks were updated accordingly.

To decently review this PR, perhaps easier to run through the notebook in VSCode...


notebooks/resources/MOOC_refdata_RDM_exploration.png
kvantricht, line 1: - Introduction could get a section on how to run the notebook.
- `Before you start` does not mention restricted yet?
- Should we mention in the notebook that next to this interaction with RDM, they can also browse interactively?
- Sect. 2.2 has a type "intoto"
- Visualizing the resulting geoparquet: this is super cool! (have you tested if this works correctly on Terrascope already?)
- The bbox drawer for filtering on extent still shows logging on "processing extent". This logging message should be made more generic because this application is now not on processing. The method is also called "get_processing_extent". Do we need to update this, e.g. just to `get_extent`?
- Sect. 3.4: when mentioning arguments in a markdown cell, you can wrap them around "``" signs to make them really look like arguments: `include_private=False`. May apply to other cells in the notebook too (Sect. 5 for sure).
- For private collections I followed the device login with wcextractions username and didn't get any hits. Is that expected? Maybe state the obvious in the description of this step that it's of course a requirement to have uploaded private collections first (and maybe link to the relevant part of the MOOC or documentation where this is explained).
- How useful is the step of `counts = rdm.get_crop_counts(ref_ids)` if you just get the codes? Can't we by default use the human readable format for this? Or add an optional argument that would allow this immediate translation?
- "Below we provide one example of this functionality, but feel free to play around and request samples for your specific area/period/crop types of interest!" The example does require the user to draw an extent? So no deterministic results. Alternatively you could use here a predefined SpatialExtent?
- Last markdown cell: can we add some next steps or call to action? Refer them to upload procedure again? Or RDM UI in general? Or the classification system? Just something to keep them going.

notebooks/notebook_utils/classifier.py
```diff
@@ -0,0 +1,501 @@
...
+    ]
+    vbox = widgets.VBox(
+        checkbox_widgets,
+        layout=widgets.Layout(width="50%", display="inline-flex", flex_flow="row wrap"),
+    )
+
+    return vbox, checkbox_widgets
+
+
+def pick_croptypes(df: pd.DataFrame, samples_threshold: int = 100):
```
kvantricht, line 65: A bit confusing that we now have this function and also a CroptypePicker for another tool. This is completely separate?

notebooks/worldcereal_RDM_demo.ipynb
kvantricht, line 1: By accident commented on the PNG file. So there you'll find my feedback on the notebook.

src/worldcereal/rdm_api/rdm_collection.py
```diff
@@ -45,3 +52,89 @@ def __init__(self, **metadata):
...
+            scroll_wheel_zoom=True,
+        )
+
+        # Add the rectangle to the map
+        m.add_layer(rectangle)
+
+        return m
+
+
+def crop_stats_from_metadata(
```
kvantricht, line 88: any particular reason for this to be a separate method in this module and not a method of the RDMInteraction class in the other one? In which case you could get crop stats and it would internally call the metadata query and then this method. Just a thought, nothing crucial though.

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -491,3 +504,162 @@ def download_samples(
...
+        outfile = Path(dst_path) / f"{ref_id}_metadata.xlsx"
+        Path(dst_path).mkdir(parents=True, exist_ok=True)
+        with open(outfile, "wb") as f:
+            f.write(response.content)
+
+        logger.info(f"Metadata for collection {ref_id} downloaded to {dst_path}")
+
+        return str(outfile)
+
+    def download_collection_samples(
```
kvantricht, line 569: I'm a bit confused by this method on the one hand and `download_samples` on the other. The latter one returns all samples in-memory as GeoDataFrame, while the former one basically downloads the original geoparquet file from RDM. This may have to be clarified a bit more, optionally also by changing the method names.

This one e.g. could be renamed to something like `download_collection_geoparquet` or `download_collection`,
while the latter could in fact be something like `get_samples`? That may be also more transparent in terms of one method downloading a file and the other basically constructing a GeoDataFrame.

src/worldcereal/utils/map.py
kvantricht, line 1: we need to make sure that we instruct users who want to install locally do in fact install with the option that brings in things like `ipyleaflet`. `[project.optional-dependencies]` could get a section on `notebooks` for example. Then if users do a `pip install .[notebooks]`, it will also install the requirements for the notebooks.

notebooks/notebook_utils/classifier.py
jdegerickx: I agree.
This function is specifically for the phase I extractions. Removing it now will cause the v1 notebooks not to work anymore.
Idea is to replace this by the CroptypePicker in version 2 of the system.
I also feel like we should remove/update the v1 notebooks before the release of v2, so we can remove outdated functionality like this one...

src/worldcereal/utils/map.py
kvantricht: Yes you can add it as a third point in the Terrascope option, and also update the pip install instruction to include the train and notebooks options.

notebooks/worldcereal_RDM_demo.ipynb
```diff
@@ -76,11 +78,18 @@
...
     "### Before you start\n",
     "\n",
-    "Reference datasets (termed \"collections\") in the WorldCereal RDM can either be open to the public, or private (only accessible to the user who uploaded the data). \n",
+    "Reference data in the WorldCereal RDM are organized in collections.<br>\n",
+    "A collection contains observations derived from a single source and for a single year.<br>\n",
+    "\n",
+    "Each collection is characterized by a **data privacy level**, controlling who can access the data:<br>\n",
+    "- *Public datasets* can be accessed by anyone and have been explicitly curated by the WorldCereal consortium;\n",
+    "- *Private datasets* can only be accessed by the user who uploaded the data;\n",
+    "- *Restricted datasets* are private datasets which have been shared with the WorldCereal consortium, but not with the general public.\n",
```
kvantricht, line 87: "which were approved by the uploader to be used for training the global WorldCereal, but which are not shared publicly".

kvantricht (approved):
Added one final comment, once addressed, this branch is ready to be merged. Immediately afterwards, I'll bump the version number on main.

kvantricht (approved):
Looks good!

PR #275, opened 2025-01-28
Do not use `spatial_extent` for `load_stac`

kvantricht:
By removing `spatial_extent` in calls to `load_stac`, we can circumvent the issue that occurs when running single-point jobs that would trigger an error in `load_stac`. Tested both for single-point jobs as well as for ordinary patch extractions.


scripts/extractions/extract.py
```diff
@@ -388,12 +392,30 @@ def run_extractions(
     tracking_df_path = output_folder / "job_tracking.csv"
 
     # Load the input dataframe and build the job dataframe
-    input_df = load_dataframe(input_df)
+    input_gdf = load_dataframe(input_df)
+
+    # Load the already extracted data and remove sample_ids that have already been extracted
+    if skip_existing:
```
VincentVerelst, line 398: Isn't this only relevant for point extractions on private reference data? Maybe this is something we can move to `prepare_job_dataframe` for point extractions in that case. That way the resulting job database (i.e. the CSV file) will only contain the relevant jobs

kvantricht: Good point. Not just for private reference data, because I actually used this for the huge point extractions too, to be able to restart fresh with a new job database, skipping samples that were already done. But it's anyway in the wrong place here, because this check runs even when a job_tracking.csv is already there. But where exactly would you put it then? I could imagine at some point it could be useful for patches as well. If you need start all the way over, resetting job database, but you don't want to launch for patches that were fully done already.

VincentVerelst: If this is a bespoke use case, I would leave it up to the user to account for this when creating their `pandas.DataFrame` for job management at the very start of the pipeline. 

If this is something that happens quite often in WorldCereal we can pre-define a specific `prepare_job_dataframe` helper function (like the ones we already have for normal patch and point extractions). We can even encorporate this in the existing `prepare_job_dataframe` functions we already have, by adding the optional `skip_existing` argument. 

Once the `pandas.DataFrame` is created, the high-level pipeline looks as follows:

```python
from openeo.extra.job_management import CsvJobDatabase, MultiBackendJobManager

# Create a new job database based on user-defined, or pre-defined WorldCereal code
df = prepare_job_dataframe(...)

# Initialize your job database
job_db = CsvJobDatabase('/path/to/job_tracker.csv').initialize_from_df(df)

# pass your job database to the MultiBackendJobManager
job_manager = MultiBackendJobManager(...)
job_manager.run_jobs(..., job_db=job_db)


```

kvantricht: To keep going, I suggest to leave this part of the code out and keep that for a larger upgrade to incorporate the CsvJobDatabase stuff and official JobManager. What do you think?

VincentVerelst: Ok sounds good!

VincentVerelst (approved):

PR #278, opened 2025-02-05
Update patch post job actions

VincentVerelst:
Post job actions for patch extractions have been updated so that:

- The resulting STAC items now contain the same metadata in their properties as the underlying netCDFs in their attributes
- The S2 tile is assigned based on the latest version of GFMap
- As much as possible metadata follows the schema of existing STAC extensions
- Users can now disable `write_stac_api` for S1 and S2 patch extractions; useful in testing

Note that:
- `epsg` had to be removed from `generate_output_path_patch`, since - with the new version of `split_job_s2grid` - there is no way to find out the EPSG code, without accessing the STAC metadata. This STAC metadata is not accessed by `generate_output_path_patch`, so would require a rewriting of the `GFMapJobManager`

The extractions have been run on a test dataset and were successfully loaded back in openEO using `load_stac`.

### DIFF
--- a/src/worldcereal/extract/common.py
+++ b/src/worldcereal/extract/common.py
@@ -110,7 +110,7 @@ def post_job_action_patch(
             "start_date": row.start_date,
             "end_date": row.end_date,
             "valid_time": valid_time,
-            "GFMAP_version": version("openeo_gfmap"),
+            "processing:version": version("openeo_gfmap"),
             "creation_date": datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
             "description": description,
             "title": title,
@@ -123,7 +123,7 @@ def post_job_action_patch(
         }
 
         if s1_orbit_fix:
-            new_attributes["orbit_state"] = row.orbit_state
+            new_attributes["sat:orbit_state"] = row.orbit_state
             item.id = item.id.replace(".nc", f"_{row.orbit_state}.nc")
 
         # Saves the new attributes in the netcdf file
--- a/src/worldcereal/extract/common.py
+++ b/src/worldcereal/extract/common.py
@@ -195,13 +195,12 @@ def generate_output_path_patch(
 
     s2_tile_id = row.s2_tile
     utm_zone = str(s2_tile_id[0:2])
-    epsg = s2_grid[s2_grid.tile == s2_tile_id].iloc[0].epsg
 
     subfolder = root_folder / ref_id / utm_zone / s2_tile_id / sample_id
 
     return (
         subfolder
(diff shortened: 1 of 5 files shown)


src/worldcereal/extract/common.py
```diff
@@ -123,7 +124,7 @@ def post_job_action_patch(
         }
 
         if s1_orbit_fix:
```
kvantricht, line 126: I don't recall the original aim of this boolean. What happens if it's false?

src/worldcereal/stac/stac_api_interaction.py
```diff
@@ -67,7 +67,9 @@ def __init__(
             )
         self.sensor = sensor
         self.base_url = base_url
-        self.collection_id = f"worldcereal_{sensor.lower()}_patch_extractions"
+        self.collection_id = (
+            f"worldcereal_{sensor[:-1].lower()}_{sensor[-1]}_patch_extractions"
```
kvantricht, line 71: is this future proof for different kinds of sensors?

kvantricht (commented):
Already looking good. With respect to epsg comment, we were only using it to add to the properties?

src/worldcereal/extract/common.py
VincentVerelst: I also didn't write this part of the code. It looks like it just adds the S1 orbit state to the item ID and filename. This is necessary to make it unique, since we extract both ascending and descending.

src/worldcereal/stac/stac_api_interaction.py
VincentVerelst: No, currently the STAC API interaction is only designed for sentinel1 and 2: https://github.com/WorldCereal/worldcereal-classification/blob/239-patch-post-job-actions/src/worldcereal/stac/stac_api_interaction.py#L64

We could make a dicitionary catalog, to map sensor IDs to STAC collection IDs?

VincentVerelst:
> Already looking good. With respect to epsg comment, we were only using it to add to the properties?

The EPSG is already part of the item properties straight from openeo. We were also using EPSG in the output path generation, to make it a specific subfolder. However, with the current setup of GFMap, there's no way for us to access the EPSG code used by openeo in the output path generator.

src/worldcereal/stac/stac_api_interaction.py
kvantricht: Good idea, this way at least we'll get an error as soon as we'd try a new sensor that doesn't have a collection yet.

kvantricht (approved):

PR #284, opened 2025-02-11
added worldcereal_crop_extent and type UDP to the repo

VincentVerelst:
Just added the two JSON files of the `worldcereal_crop_type` and `wordcereal_crop_extent` UDPs to this repo. 
Afterwards will create a PR in the APEx repo to remove the UDPs there and update the benchmark scenarios.

### DIFF
--- a/src/worldcereal/udp/worldcereal_crop_extent.json
+++ b/src/worldcereal/udp/worldcereal_crop_extent.json
@@ -1290,7 +1290,7 @@
       "optional": true
     }
   ],
-  "minimal_job_options": {
+  "default_job_options": {
     "driver-memory": "4g",
     "executor-memory": "2g",
     "executor-memoryOverhead": "1g",
--- a/src/worldcereal/udp/worldcereal_crop_extent.json
+++ b/src/worldcereal/udp/worldcereal_crop_extent.json
@@ -247,20 +247,6 @@
         },
         "id": "SENTINEL1_GRD",
         "properties": {
-          "sat:orbit_state": {
-            "process_graph": {
-              "eq1": {
-                "process_id": "eq",
-                "arguments": {
-                  "x": {
-                    "from_parameter": "value"
-                  },
-                  "y": "DESCENDING"
-                },
-                "result": true
-              }
-            }
-          },
           "polarisation": {
             "process_graph": {
               "eq2": {
--- a/src/worldcereal/udp/worldcereal_crop_extent.json
+++ b/src/worldcereal/udp/worldcereal_crop_extent.json
@@ -1193,6 +1193,18 @@
         }
       },
       "result": true
+    },
+    "saveresult2": {
+      "process_id": "save_result",
+      "arguments": {
+        "data": {
+          "from_node": "apply5"
+        },
+        "format": "GTiff",
+        "options": {
+          "filename_prefix": "worldcereal-cropland-extent-postprocessed"
+        }
+      }
     }
   },
   "id": "worldcereal_crop_extent",
--- a/src/worldcereal/udp/worldcereal_crop_extent.json
+++ b/src/worldcereal/udp/worldcereal_crop_extent.json
@@ -1,0 +1,1317 @@
+{
+  "process_graph": {
+    "loadcollection1": {
+      "process_id": "load_collection",
+      "arguments": {
+        "bands": [
+          "B02",
+          "B03",
+          "B04",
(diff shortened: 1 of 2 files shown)


kvantricht (commented):
From this point onward, we have to take special care of versioning. So my question now is, once we merge this PR, apex will get the UDP here tied to a specific version. What's the procedure here? We need to bump version on main once this is merged, tag it, and use that version tag to inform apex on current UDP version? So any subsequent change on main does not directly affect apex? And when we want to release a new version, we need a new tag and communicate this to apex?

VincentVerelst:
@kvantricht , there are two options:

- We provide the raw github URL to the UDP on the worldcereal-classification main branch to APEx. This way, when the UDP is updated on the main branch, it will automatically be updated for APEx, and also for users who are calling the UDP with this raw github URL. Here is an example of the benchmark scenario in APEx using this raw URL: https://github.com/ESA-APEx/apex_algorithms/blob/main/benchmark_scenarios/worldcereal_crop_type.json#L10 
NOTE: the URL is currently pointing to the APEx main branch, because the UDPs still need to be moved to worldcereal-classification

- Instead, we can also provide the raw github URL linked to a certain commit (instead of linked to a certain branch). Like for example https://github.com/ESA-APEx/apex_algorithms/commit/d4143334e07f22242a0e529636959f0e71d45348
Changes to the main branch will not affect this UDP, ensuring that it's always the same. However, whenever we want to update the UDP, we now also have to update the URL in APEx, and also make sure users are using the new URL in their namespace

To summarize:
- Option1: Users and APEx repo will always have the latest version of the UDP. Where latest version = current version on worldcereal-classification main branch
- Option2: Users and APEx will always have a static version of the UDP. Users will have to manually update their namespace in order to get the latest version of the UDP

VincentVerelst:
Using tags can also work I think. You can also get raw links to a certain tagged version.

kvantricht:
Thanks for the explanation @VincentVerelst ! But if I recall @jdries 's comment in previous openEO meeting a while ago, option 2 would be the preferred one from APEx point of view to ensure stability as long as no deliberate update of the UDP version is pushed to APEx? Or am I mistaken?

VincentVerelst:
> Thanks for the explanation @VincentVerelst ! But if I recall @jdries 's comment in previous openEO meeting a while ago, option 2 would be the preferred one from APEx point of view to ensure stability as long as no deliberate update of the UDP version is pushed to APEx? Or am I mistaken?

When using option 1, you indeed might run into the problem that the current info, or current benchmark scenario in APEx is not up-to-date with the latest version of the UDP on worldcereal-classification. @jdries might be better suited to advise us here.

jdries:
My idea was to work with tags, so not pointing to main branch directly. This is under the assumption that main branch can sometimes contain 'unstable' udp's that are still under development. 
By using tags, you can safely merge in pull requests, without having to worry about user impact.

The idea is then that you really do an explicit 'release'. With git tags, it might also be an option to have a kind of fixed tag, that you point to different commits when you do releases, circumventing the need to continuously updating apex repo.
 
Semantic versioning is also an option:
git_tag_1 -> point to latest major version 1
git_tag_1.0.3 -> points to a specific feature/bugfix version 
git_tag_unstable -> latest from main
git_tag_2 -> points to latest major version 2, indicating that there's incompatibilities with major version 1

VincentVerelst:
Thanks @jdries! So @kvantricht, since we'll be hosting multiple UDPs, I believe it's best to create one tag name per UDP? E.g.:
- worldcereal_crop_type_1.0.0
- worldcereal_crop_type_1.0.3
- worldcereal_crop_extent_1.0.0

kvantricht:
for my understanding, this is still a regular git tag to a specific commit in the repo? So at the moment we have the typical v2.0.1 tags e.g. which we could keep to release system updates, but separately we'd have tags with a naming convention to organize specific repo status pointers for the UDPs?

VincentVerelst:
> for my understanding, this is still a regular git tag to a specific commit in the repo? So at the moment we have the typical v2.0.1 tags e.g. which we could keep to release system updates, but separately we'd have tags with a naming convention to organize specific repo status pointers for the UDPs?

That is correct

kvantricht:
Then it sounds good to me!

kvantricht (approved):
Nice, thanks! We should notify after merge to Gisat that these UDPs are now here.

PR #289, opened 2025-02-20
connect to CDSE with correct URL

jdegerickx:
simplified code in generate_map that takes care of connecting to openeo backend


jdegerickx:
@kvantricht, Verified with Jeroen Dries.
According to him, this was indeed due to a potential firewall issue.
But if one representative job succeeds, there is likely no longer a problem.
Did a default cropland test run and that one succeeded, so I guess this is safe to commit...

kvantricht (approved):
Thanks for checking!

PR #290, opened 2025-02-21
Dwnld notebook

jdegerickx:
New notebook showcasing download of 2021 products using Federation openeo backend.

Implemented new functionality to request seasonal extent from phase I AEZ file.

Users can choose to download all products, or a specific season.


jdegerickx:
I don't understand the linting error.
openeo is imported in job.py...

PR #292, opened 2025-03-03
move extraction functions scripts > src

jdegerickx:
Core functions previously in scripts > extract.py have been moved to src > extractions > common.py

Functionality has been split into 3 "stages" (prepare extractions, run extractions and merge extractions).
This has been deliberately done to support running extractions using push notifications in scripts > extract.py. The way in which the extractions are fired up there, has not been changed.

Some key consequences:
- some common util functions have been moved from common.py to src > extractions > utils.py to avoid circular imports
- default job options have been defined for each type of extractions in the respective files (e.g. DEFAULT_JOB_OPTIONS_POINT_WORLDCEREAL).
- instead of having 3 separate arguments (memory, python_memory and max_executors), I propose to have a "custom_job_options" option for more advanced users that want to change specific job options.

### DIFF
--- a/scripts/extractions/extract.py
+++ b/scripts/extractions/extract.py
@@ -167,13 +167,13 @@ def run_extractions(
     """
 
     # Compile custom job options
-    custom_job_options: Dict[str, Union[str, int]] = {}
+    job_options: Dict[str, Union[str, int]] = {}
     if memory:
-        custom_job_options["memory"] = memory
+        job_options["memory"] = memory
     if python_memory:
-        custom_job_options["python_memory"] = python_memory
+        job_options["python_memory"] = python_memory
     if max_executors:
-        custom_job_options["max_executors"] = max_executors
+        job_options["max_executors"] = max_executors
 
     # Prepare extraction jobs
     job_manager, job_df, datacube_fn, tracking_df_path = prepare_extraction_jobs(
@@ -181,7 +181,7 @@ def run_extractions(
         output_folder,
         samples_df_path,
         max_locations_per_job=max_locations_per_job,
-        custom_job_options=custom_job_options,
+        job_options=job_options,
         parallel_jobs=parallel_jobs,
         restart_failed=restart_failed,
         extract_value=extract_value,
--- a/scripts/extractions/extract.py
+++ b/scripts/extractions/extract.py
@@ -2,120 +2,16 @@
 own functions, but the setup and main thread execution is done here."""
 
 import argparse
-import os
-from datetime import datetime
 from pathlib import Path
-from typing import Callable, Dict, Optional, Union
+from typing import Dict, Optional, Union
 
-import geopandas as gpd
-import pandas as pd
-import requests
-from openeo.rest import OpenEoApiError, OpenEoApiPlainError, OpenEoRestError
 from openeo_gfmap import Backend
-from openeo_gfmap.manager.job_manager import GFMAPJobManager
 
-from worldcereal.extract.common import (
-    _merge_extraction_jobs,
-    _prepare_extraction_jobs,
-    pipeline_log,
-)
+from worldcereal.extract.common import run_extractions
 from worldcereal.stac.constants import ExtractionCollection
 
-# Pushover API endpoint, allowing to send notifications to personal devices.
-PUSHOVER_API_ENDPOINT = "https://api.pushover.net/1/messages.json"
 
-
-def send_notification(message: str, title: str = "OpenEO-GFMAP") -> None:
-    """Send a notification to the user's device using the Pushover API.
-
-    The environment needs to have the PUSHOVER_USER_TOKEN and PUSHOVER_APP_TOKEN
-    variables setup.
-    """
-    user_token = os.getenv("PUSHOVER_USER_TOKEN")
-    app_token = os.getenv("PUSHOVER_APP_TOKEN")
-
-    if user_token is None or app_token is None:
-        pipeline_log.warning("No pushover tokens found, skipping the notification.")
-        return
-
-    data = {
-        "token": app_token,
-        "user": user_token,
-        "message": message,
-        "title": title,
-    }
-    response = requests.post(PUSHOVER_API_ENDPOINT, data=data)
-
-    if response.status_code != 200:
-        pipeline_log.error("Error sending the notification: %s", response.text)
-
-
-def manager_main_loop(
-    manager: GFMAPJobManager,
-    collection: ExtractionCollection,
-    job_df: gpd.GeoDataFrame,
-    datacube_fn: Callable,
-    tracking_df_path: Path,
-) -> None:
-    """Main loop for the job manager, re-running it whenever an uncatched
-    OpenEO exception occurs, and notifying the user through the Pushover API
-    whenever the extraction start or an error occurs.
-    """
-    latest_exception_time = None
-    exception_counter = 0
-
-    while True:
-        pipeline_log.info("Launching the jobs manager.")
-        try:
-            send_notification(
-                title=f"WorldCereal Extraction {collection.value} - Started",
-                message="Extractions have been started.",
-            )
-            manager.run_jobs(job_df, datacube_fn, tracking_df_path)
-            return
-        except (
-            OpenEoApiPlainError,
-            OpenEoApiError,
-            OpenEoRestError,
-            requests.exceptions.ChunkedEncodingError,
-            requests.exceptions.HTTPError,
-        ) as e:
-            pipeline_log.exception("An error occurred during the extraction.\n%s", e)
-            send_notification(
-                title=f"WorldCereal Extraction {collection.value} - OpenEo Exception",
-                message=f"An OpenEO/Artifactory error occurred during the extraction.\n{e}",
-            )
-            if latest_exception_time is None:
-                latest_exception_time = pd.Timestamp.now()
-                exception_counter += 1
-            # 30 minutes between each exception
-            elif (datetime.now() - latest_exception_time).seconds < 1800:
-                exception_counter += 1
-            else:
-                latest_exception_time = None
-                exception_counter = 0
-
-            if exception_counter >= 3:
-                pipeline_log.error(
-                    "Too many OpenEO exceptions occurred in a short amount of time, stopping the extraction..."
-                )
-                send_notification(
-                    title=f"WorldCereal Extraction {collection.value} - Failed",
-                    message="Too many OpenEO exceptions occurred, stopping the extraction.",
-                )
-                raise e
-        except Exception as e:
-            pipeline_log.exception(
-                "An unexpected error occurred during the extraction.\n%s", e
-            )
-            send_notification(
-                title=f"WorldCereal Extraction {collection.value} - Failed",
-                message=f"An unexpected error occurred during the extraction.\n{e}",
-            )
-            raise e
-
-
-def run_extractions(
+def main(
     collection: ExtractionCollection,
     output_folder: Path,
     samples_df_path: Path,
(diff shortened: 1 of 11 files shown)


src/worldcereal/extract/patch_meteo.py
```diff
@@ -26,10 +34,9 @@ def create_job_patch_meteo(
     connection: openeo.DataCube,
     provider,
     connection_provider,
-    executor_memory: str,
-    python_memory: str,
-    max_executors: int,
+    custom_job_options: Optional[Dict[str, Union[str, int]]] = None,
```
VincentVerelst, line 37: In light of keeping the naming in line with the openEO spec; could we just name this kwarg `job_options`?

src/worldcereal/extract/patch_s1.py
```diff
@@ -122,9 +130,7 @@ def create_job_patch_s1(
     connection: openeo.DataCube,
     provider,
     connection_provider,
-    executor_memory: str,
-    python_memory: str,
-    max_executors: int,
+    custom_job_options: Optional[Dict[str, Union[str, int]]] = None,
```
VincentVerelst, line 133: Same comment as above on the naming, consider `job_options`?

src/worldcereal/extract/patch_s2.py
```diff
@@ -72,10 +88,9 @@ def create_job_patch_s2(
     connection: openeo.DataCube,
     provider,
     connection_provider,
-    executor_memory: str,
-    python_memory: str,
-    max_executors: int,
+    custom_job_options: Optional[Dict[str, Union[str, int]]] = None,
```
VincentVerelst, line 91: Same comment as above on the naming, consider `job_options`?

src/worldcereal/extract/patch_worldcereal.py
```diff
@@ -124,9 +134,7 @@ def create_job_patch_worldcereal(
     connection: openeo.DataCube,
     provider,
     connection_provider,
-    executor_memory: str,
-    python_memory: str,
-    max_executors: int,
+    custom_job_options: Optional[Dict[str, Union[str, int]]] = None,
```
VincentVerelst, line 137: Same comment as above on the naming, consider `job_options`?

src/worldcereal/extract/common.py
```diff
@@ -204,85 +212,327 @@ def generate_output_path_patch(
...
+    if collection == ExtractionCollection.POINT_WORLDCEREAL:
+        pipeline_log.info("Merging Geoparquet results...")
+        ref_id = Path(samples_df_path).stem
+        merge_output_files_point_worldcereal(output_folder=output_folder, ref_id=ref_id)
+        pipeline_log.info("Geoparquet results merged successfully.")
+
+    return
+
+
+def run_extractions(
```
VincentVerelst, line 469: If I understand correctly, this will be the only function called by users? Maybe we could consider making all other functions (`prepare_extraction_jobs`, `run_extraction_jobs`, `merge_extraction_jobs)` private?

src/worldcereal/extract/patch_s1.py
jdegerickx: and in the actual passing of job options, I should do something like:

# Set job options
   final_ job_options = copy.deepcopy(DEFAULT_JOB_OPTIONS_PATCH_METEO)
    if job_options:
        final_job_options.update(job_options)

    return cube.create_job(
        out_format="NetCDF",
        title=f"GFMAP_Extraction_AGERA5_{h3index}_{valid_time}",
        sample_by_feature=True,
        job_options=final_job_options,
    )

?

src/worldcereal/extract/common.py
jdegerickx: Yeah guess so.
This means that if a user wants to re-initiate extractions, he always needs to call the run_extractions function and that function will again construct datacube_fn, post_job_action_fn etc. But I guess these are not really time-consuming steps so it shouldn't impact the user experience, right?
Or do you see a need for being able to directly call run_extraction_jobs function?

src/worldcereal/extract/common.py
```diff
@@ -204,85 +212,327 @@ def generate_output_path_patch(
...
+    post_job_actions = {
+        ExtractionCollection.PATCH_SENTINEL1: partial(
+            post_job_action_patch,
+            extract_value=extract_value,
+            description="Sentinel-1 GRD backscatter observations, processed with Orfeo toolbox.",
+            title="Sentinel-1 GRD",
+            spatial_resolution="20m",
+            s1_orbit_fix=True,
+            sensor="Sentinel1",
+            write_stac_api=write_stac_api,
```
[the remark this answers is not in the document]
jdegerickx: I agree to have this set to False by default everywhere and only manually set to True if we do extractions ourselves

src/worldcereal/extract/common.py
VincentVerelst: No, I didn't think of that. In that case it does make sense to expose all three functions as public to the user. Unless we provide something in the `run_extractions` job that detects whether a user restarts an existing extraction or starts a new one. In case of the former, the `prepare_extraction_jobs` step can be skipped and the `GFMapJobManager` needs to run with `restart_failed=True`.

src/worldcereal/extract/patch_s1.py
VincentVerelst: I think the implementation makes sense in this way. Users can enter `job_options` which will override the default ones. It was just a matter of semantics, i.e. renaming the argument :)

jdegerickx: adjusted naming in most recent commit

src/worldcereal/extract/common.py
jdegerickx: added some logic to deal with existing job tracking df. If it exists, it just gets loaded instead of re-created.

scripts/extractions/extract.py
```diff
@@ -339,16 +118,16 @@ def manager_main_loop(
 def run_extractions(
```
VincentVerelst, line 118: Is there a reason we can't just use the `run_extractions` function from `src`?

jdegerickx: yes, because the push notification is not implemented in src and is probably not something that we want in src?

jdegerickx: guess the notifications are still useful to have for large batches of extractions.

Instead of only wrapping the running of the jobs in such a notification function, I could wrap the entire run_extractions function into it...
This would avoid having to redefine the run_extractions function in scripts...

jdegerickx: but maybe this is something for @kvantricht to have a look at, as he has been using this script.
For me, would be fine to keep as is...

scripts/extractions/extract.py
```diff
@@ -3,56 +3,24 @@
...
 import requests
 from openeo.rest import OpenEoApiError, OpenEoApiPlainError, OpenEoRestError
 from openeo_gfmap import Backend
-from openeo_gfmap.backend import cdse_connection
 from openeo_gfmap.manager.job_manager import GFMAPJobManager
-from openeo_gfmap.manager.job_splitters import load_s2_grid, split_job_s2grid
 
 from worldcereal.extract.common import (
-    generate_output_path_patch,
+    _merge_extraction_jobs,
```
kvantricht, line 18: Having them defined as private functions, but then importing them outside `src` should probably be avoided. But let's first see if we really need the duplication. It would be preferrable to not have things duplicated. Maybe we can afterwards still implement optional notification using e.g. decorators.

src/worldcereal/extract/common.py
```diff
@@ -1,49 +1,58 @@
...
+)
+from worldcereal.extract.patch_s2 import (
+    create_job_dataframe_patch_s2,
+    create_job_patch_s2,
+)
+from worldcereal.extract.point_worldcereal import (
+    create_job_dataframe_point_worldcereal,
+    create_job_point_worldcereal,
+    generate_output_path_point_worldcereal,
+    merge_output_files_point_worldcereal,
```
kvantricht, line 35: is this always product specific? Cannot be made generic? We are importing more and more small methods for each product category which is quickly becoming very hard to manage.

src/worldcereal/extract/common.py
```diff
@@ -1,49 +1,58 @@
...
-stream_handler.addFilter(ManagerLoggerFilter())
+from worldcereal.extract.patch_worldcereal import (  # isort: skip
+    create_job_patch_worldcereal,
+    create_job_dataframe_patch_worldcereal,
+    post_job_action_patch_worldcereal,
+    generate_output_path_patch_worldcereal,
+)
 
 
 def post_job_action_patch(
```
kvantricht, line 58: @VincentVerelst we now have things like WorldCereal point extractions which has all methods defined inside it. Then for the patches we define post job actions and path generator here in `common`. Isn't this becoming a bit complicated to keep the overview? How future-proof is this for future product categories? And how easy or complicated will it be to use WorldCereal extraction methods on your own defined product?

src/worldcereal/extract/common.py
```diff
@@ -204,85 +212,371 @@ def generate_output_path_patch(
...
+
+    # Filter the input dataframe to only keep the locations to extract
+    samples_gdf = samples_gdf[samples_gdf["extract"] >= extract_value].copy()
+
+    # Split the locations into chunks of max_locations
+    split_dfs = []
+    pipeline_log.info(
+        "Performing splitting by the year...",
+    )
+    samples_gdf["valid_time"] = pd.to_datetime(samples_gdf.valid_time)
```
kvantricht, line 244: Are we already doing a check on required attributes somewhere before even starting the extractions, so we don't end up with a result which is incompatible?

scripts/extractions/extract.py
jdegerickx: @kvantricht, thanks for the comment! Let me then implement the retry logic in src and get rid of the push notifications.
At some point you mentioned you were going to add upon this retry logic, or am I mistaken?
In any case, let me know if there are specific changes needed based on your experience in running extractions!

kvantricht (commented):
Next to the outstanding comments, has the refactored workflow been successfully tested already?

scripts/extractions/extract.py
kvantricht: yeah actually the way the retry is now implemented is a bit complicated and doesn't always work that well. We used to work with the `retry` package which is perfectly fine, but then it's also yet another dependency. Maybe the golden bullet here is what we actually did in satio, which is taking core functionality of `retry` and add it to the package itself (https://git.vito.be/projects/LCLU/repos/satio/browse/satio/utils/retry.py). So we can basically copy this `retry.py` from satio to `src/utils/retry.py` in this package and then import its functionality wherever we need it. Then we can use the retry decorator nicely with the extraction loop. How does that sound?

src/worldcereal/extract/common.py
jdegerickx: In the workflow I'm preparing for the notebook, this attribute check is done during preparation of the sample dataframe, but is not part of the extractions workflow itself.
Would you rather have me to implement it directly as part of the extractions?

kvantricht: @VincentVerelst wasn't this something you suggested? I don't have a strong opinion.

scripts/extractions/extract.py
jdegerickx: added retry.py.
At the very start, it tries to import the decorator functionality, but that resulted in a mypy error, stating that an additional package was required. I avoided that by implementing by default the fall-back mechanism that was already there in satio

@kvantricht, please quickly check whether you agree with the default parameters (number of retries etc.) and whether it's ok to have possibility to set it as env variable (as was done in satio).

scripts/extractions/extract.py
jdegerickx: tackled by importing run_extractions function from src

src/worldcereal/extract/common.py
jdegerickx: suggest to leave as is for now in order to proceed...

scripts/extractions/extract.py
kvantricht: Looks good; I didn't check in detail how satio dealt with it in the past, but the aim was indeed to not import from retry package anyway. For the env variables, maybe rather call them something like `WORLDCEREAL_EXTRACTION_RETRIES` to have it specifically set for the extractions?

kvantricht: For the defaults, it looks like these are much shorter retries and less attempts than the manually implemented ones. Maybe we should be a bit closer to those? 5 retries, more initial time between attempts (e.g. 10s) and a backoff (multiplier of the delay for each attempt) of 5?

jdegerickx: thanks! pushed the requested changes

kvantricht (approved):
let's proceed for now

PR #297, opened 2025-03-07
allow visualization of multiple collections on the map

jdegerickx:
previously we had a function that visualizes the spatial extent of one collection on a map.
I extended this function to allow more than one collection to be visualized on the same map, which is more convenient if you want to check data availability in RDM!


kvantricht (approved):
Looks good!

PR #298, opened 2025-03-10
Support `BaseGeometry` as input for `RdmInteraction.get_samples()`

VincentVerelst:
- The `bbox` argument for  `RdmInteraction.get_samples()` has been renamed to `spatial_extent`
- `spatial_extent` can now both be a `BoundingBoxExtent` or a `BaseGeometry`
- In the SQL query to the RDM an extra `ST_Simplify` has been added to the geometries to avoid `TopologyException`

### DIFF
--- a/src/worldcereal/rdm_api/rdm_interaction.py
+++ b/src/worldcereal/rdm_api/rdm_interaction.py
@@ -17,7 +17,7 @@ from openeo.rest.auth.oidc import (
 from openeo_gfmap import BoundingBoxExtent, TemporalContext
 from requests.adapters import HTTPAdapter
 from shapely import wkb
-from shapely.geometry import Point, Polygon
+from shapely.geometry import MultiPolygon, Polygon
 from shapely.geometry.base import BaseGeometry
 from urllib3.util.retry import Retry
 
@@ -110,7 +110,9 @@ class RdmInteraction:
 
     def get_collections(
         self,
-        spatial_extent: Optional[Union[BoundingBoxExtent, BaseGeometry]] = None,
+        spatial_extent: Optional[
+            Union[BoundingBoxExtent, Polygon, MultiPolygon]
+        ] = None,
         temporal_extent: Optional[TemporalContext] = None,
         include_public: Optional[bool] = True,
         include_private: Optional[bool] = False,
@@ -121,8 +123,8 @@ class RdmInteraction:
 
         Parameters
         ----------
-        spatial_extent : Optional[Union[BoundingBoxExtent, BaseGeometry]], optional
-            A user-defined bounding box, or geometry for which all intersecting collections need to be found.
+        spatial_extent : Optional[Union[BoundingBoxExtent, Polygon, MultiPolygon]], optional
+            A user-defined bounding box, or shapely Polygon or MultiPolygon for which all intersecting collections need to be found.
             CRS should be EPSG:4326.
             If None, all available data will be queried, by default None
         temporal_extent : Optional[TemporalContext], optional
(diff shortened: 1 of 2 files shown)


VincentVerelst:
@jdegerickx , still need to rewrite some unit tests, but the code itself can already be reviewed

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -434,8 +455,8 @@ def get_samples(
             If True, only download a subset of the samples (for which extract attribute ==1)
             If False, extract all samples.
             Default is False.
-        bbox : Optional[BoundingBoxExtent], optional
-            A user-defined bounding box for which all intersecting samples need to be found.
+        spatial_extent : Optional[Union[BoundingBoxExtent, BaseGeometry]], optional
```
jdegerickx, line 458: in get_collections, we require EPSG to be 4326 and also implemented a check to verify that is indeed the case.
Now wondering whether we should have the same here as well?

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -324,8 +338,8 @@ def _setup_sql_query(
         ----------
         urls : List[str]
             A list of URLs of the GeoParquet files.
-        bbox : BoundingBoxExtent
-            A user-defined bounding box.
+        spatial_extent : Union[BoundingBoxExtent, BaseGeometry]
```
jdegerickx, line 341: should we limit the type of geometry? 
For instance, passing a single point will result in errors?

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -148,16 +149,29 @@ def get_collections(
                 self.authenticate()
 
         # Handle geometry
-        if bbox is None:
-            bbox = BoundingBoxExtent(
+        if isinstance(spatial_extent, BaseGeometry):
```
jdegerickx, line 152: here we assume the geometry is in EPSG:4326. Should we check this beforehand somehow? 
Same comment in get_samples function: do we need to check CRS of provided geometry?

jdegerickx (changes requested):
Thanks!
My main concern is on user input validation: should we add some more checks to ensure correct CRS and limit the types of supported geometries?

VincentVerelst:
@jdegerickx, I've added an `assert_valid_spatial_extent` function that checks everything you mentioned in the comments above. Happy to hear your thoughts!

src/worldcereal/rdm_api/rdm_interaction.py
```diff
@@ -765,6 +765,11 @@ def assert_valid_spatial_extent(
         """
 
         if isinstance(spatial_extent, BaseGeometry):
+            if not isinstance(spatial_extent, Point):
```
jdegerickx, line 768: I guess the "not" should be deleted? :)
Perhaps best to restrict to Polygon and MultiPolygon only?

jdegerickx (changes requested):

src/worldcereal/rdm_api/rdm_interaction.py
jdegerickx: or maybe not even multiPolygon now that I think about it? :)

VincentVerelst: no MultiPolygon needs to be supported for patch-to-point extractions :)

VincentVerelst: In theory other geometries can be supported as well. The duckdb query will just fetch all geometries from the RDM that intersect with the provided geometries. But maybe we only want to support `Polygon` and `MultiPolygon`?

jdegerickx: guess indeed those 2 make the most sense for now. I'd prefer we limit to those two to avoid unexpected behaviour...

PR #314, opened 2025-03-25
More efficient parquet loading

kvantricht:
This PR implements a filtering mechanism during parquet file loading which results in only loading rows with a `extract` value `>= extract_value`. This avoids loading first the entire parquet file (in the example of French GSAA data this even gets killed due to the high amount of features) and only then filtering for extract value.


jdegerickx (approved):
Looks fine to me!

PR #315, opened 2025-03-26
Improvements for (patch) extractions

kvantricht:
Some improvements for extractions:
- we can now use a non-default openEO image
- `ref_id` is now compulsory argument to stay in full control on this (needs an update to the notebooks!)
- permissions of patch extractions are set correctly


kvantricht:
@jdegerickx do you know by heart what updates would be needed in the notebooks for this `ref_id` argument?

VincentVerelst (approved):

jdegerickx (approved):
all good!

PR #321, opened 2025-03-28
Small patch extraction improvements

kvantricht:


src/worldcereal/extract/utils.py
```diff
@@ -73,7 +73,7 @@ def filter_extract_true(
 
 
 def upload_geoparquet_artifactory(
-    gdf: gpd.GeoDataFrame, name: str, collection: str = ""
```
VincentVerelst, line 76: You will lose backwards compatibility by making the `collection` argument mandatory

src/worldcereal/extract/patch_s1.py
```diff
@@ -149,7 +150,9 @@ def create_job_patch_s1(
 
     # Performs a buffer of 64 px around the geometry
     geometry_df = buffer_geometry(geometry, distance_m=320)
-    spatial_extent_url = upload_geoparquet_artifactory(geometry_df, row.name)
+    spatial_extent_url = upload_geoparquet_artifactory(
```
VincentVerelst, line 153: Instead of using Artifactory; we could also use the OpenEO-artifacts S3 bucket in the case of CDSE, like we do for the model deployment: https://github.com/WorldCereal/worldcereal-classification/blob/main/src/worldcereal/utils/upload.py#L98

This bucket already supports auto cleanup

src/worldcereal/extract/utils.py
kvantricht: is there another occasion where I have not changed this? Meaning that this would in fact still break something somewhere? I'm just quite reluctant on keeping this optional because it was the reason for major issues last week.

VincentVerelst: Not that I know of, but there's always a risk :)

src/worldcereal/extract/patch_s1.py
kvantricht: Nice suggestion. I made that change to this method and seems to work.

src/worldcereal/extract/utils.py
kvantricht: ok for now back to the optional kwarg

PR #322, opened 2025-03-28
ensure consistent order of probability bands

jdegerickx:
crop type inference was broken due to random ordering of probability bands.
This should fix this issue...

### DIFF
--- a/src/worldcereal/openeo/inference.py
+++ b/src/worldcereal/openeo/inference.py
@@ -29,7 +29,11 @@ class CropClassifier(ModelInference):
         return []  # Disable the dependencies from PIP install
 
     def output_labels(self) -> list:
-        class_names = self._parameters["lookup_table"].keys()
+        lut = self._parameters.get("lookup_table", None)
+        if lut is None:
+            raise ValueError("Lookup table is not defined.")
+        lut_sorted = {k: v for k, v in sorted(lut.items(), key=lambda item: item[1])}
+        class_names = lut_sorted.keys()
 
         return ["classification", "probability"] + [
             f"probability_{name}" for name in class_names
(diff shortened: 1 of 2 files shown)


src/worldcereal/openeo/inference.py
```diff
@@ -29,7 +29,11 @@ def dependencies(self) -> list:
         return []  # Disable the dependencies from PIP install
 
     def output_labels(self) -> list:
-        class_names = self._parameters["lookup_table"].keys()
+        lut = self._parameters.get("lookup_table", None)
+        if lut is None:
```
kvantricht, line 33: isn't this redundant? If you don't default to None, it will anyway raise an Exception if this key is not present. Besides, earlier on i think we already do a check on this.

src/worldcereal/openeo/inference.py
```diff
@@ -29,7 +29,11 @@ def dependencies(self) -> list:
         return []  # Disable the dependencies from PIP install
 
     def output_labels(self) -> list:
-        class_names = self._parameters["lookup_table"].keys()
+        lut = self._parameters.get("lookup_table", None)
```
kvantricht, line 32: maybe a small docstring to explain why we need the hassle of ordering?

kvantricht (changes requested):
some small comments

kvantricht:
Note that we'll need to trigger a build on Terrascope once this fix is on main.

src/worldcereal/openeo/inference.py
jdegerickx: docstrings added, both here and in predict function

kvantricht (approved):

PR #323, opened 2025-03-31
rdm auth now through CDSE

jdegerickx:
minor update to ensure rdm interaction for private datasets works again now we switched from terrascope to cdse login...


kvantricht (approved):
Don't forget to notify Mr. Geoglam ;-)

PR #332, opened 2025-04-23
Better handling of spatial filtering

kvantricht:


src/worldcereal/extract/patch_s1.py
```diff
@@ -154,8 +155,11 @@ def create_job_patch_s1(
 
     # Performs a buffer of 64 px around the geometry
     geometry_df = buffer_geometry(geometry, distance_m=320)
-    spatial_extent_url = upload_geoparquet_s3(
-        provider, geometry_df, row.name, "SENTINEL1"
+    # spatial_extent_url = upload_geoparquet_s3(
+    #     provider, geometry_df, f"{row.s2_tile}_{row.name}", "SENTINEL1"
+    # )
+    spatial_extent_url = upload_geoparquet_artifactory(
+        geometry_df, f"{row.s2_tile}_{row.name}", collection="SENTINEL2"
```
jdegerickx, line 162: why is collection here Sentinel2?

jdegerickx (commented):
would any of this have an impact to point extractions and should this be tested first?

src/worldcereal/extract/patch_s1.py
kvantricht: good catch, wrong copy paste!

src/worldcereal/openeo/preprocessing.py
```diff
@@ -141,23 +216,26 @@ def raw_datacube_S2(
...
         backend_context,
         bands=bands,
         fetch_type=fetch_type,
         **extraction_parameters,
-    )
+    ).get_cube(connection, None, temporal_extent)
 
-    return extractor.get_cube(connection, spatial_extent, temporal_extent)
+    # Do spatial filtering
+    s2_cube = spatially_filter_cube(connection, s2_cube, spatial_extent)
```
jdegerickx, line 227: this additional spatial filtering is causing issues for POINT extractions.
This should only be applied for patches, not for points. 

Same comment for the other data sources (S1, meteo...)

jdegerickx (changes requested):
Changes are currently breaking point extractions

jdegerickx (approved):

PR #346, opened 2025-05-09
327 lut crop type

VincentVerelst:
- Added a `create_inference_process_graph` function inside `generate_map` to make it much easier to generate a new UDP
- Updated `_postprocess` to no longer require `lookup_table` kwars, but instead `classifier_url` -> need to check for backwards compatibility. Since it's a private function, I assume this to be fine
- Updated the `worldcereal_crop_type` UDP itself. It should now be able to handle all kinds of `model_url`, and the UDF inside the UDP will automatically derive the LUT from the URL
- Modified `CropClassifier` class to automatically extract LUT from classifier URL


src/worldcereal/job.py
```diff
@@ -179,7 +171,14 @@ def generate_map(
             )
 
         # To use it as a mask, we need to filter out the classification band
-        cropland_mask = cropland_mask.filter_bands("classification")
+        # Use the generic 'process' to avoid client-side errors on missing metadata
+        cropland_mask = cropland_mask.process(
```
kvantricht, line 175: Why does this give problems now? It used to work before?

src/worldcereal/openeo/inference.py
```diff
@@ -36,7 +76,7 @@ def output_labels(self) -> list:
         not guarantee the order of a json object being preserved when decoding
         a process graph in the backend.
         """
-        lut = self._parameters["lookup_table"]
+        _, lut = self.load_and_prepare_model(self._parameters["classifier_url"])
```
kvantricht, line 79: How "heavy" is the call to this load and prepare? Would it make sense to try to do it once and keep both LUT and model in memory instead of disregarding the model here? OR is the lur_cache key here to avoid the overhead?

src/worldcereal/openeo/inference.py
```diff
@@ -55,17 +95,13 @@ def predict(self, features: np.ndarray) -> np.ndarray:
         import numpy as np
 
         # Classes names to codes
-        lookup_table = self._parameters.get("lookup_table", None)
+        _, lut_sorted = self.load_and_prepare_model(self._parameters["classifier_url"])
```
kvantricht, line 98: Same comment, second call just for getting the LUT but with quite some overhead?

src/worldcereal/openeo/postprocess.py
```diff
@@ -18,15 +22,51 @@ class PostProcessor(ModelInference):
...
+                f"probability_{name}" for name in lut.keys()
             ]
         return ["classification", "probability"]
 
     def dependencies(self) -> list:
         return []
 
+    @classmethod
+    @functools.lru_cache(maxsize=6)
+    def lut_from_url(cls, model_url: str) -> dict:
```
kvantricht, line 36: isn't this code duplication from `load_and_prepare_model` ? Is this openEO limitation that we cannot just centralize this once?

kvantricht (commented):
In general looking good, especially now it's verified to work. I just have some questions on the multiple calls to the model/lut loading (though caching may mean this comment is not a concern), and the duplication of the method that actually does model/lut loading.

src/worldcereal/openeo/inference.py
VincentVerelst: Tested and found no perfomance improvement in the UDP. Both benchmarks took the same amount of time and burned the same amount of credits

src/worldcereal/openeo/postprocess.py
VincentVerelst: Ah, but the method needs to be a classmethod in order for `lru_cache` to work properly

src/worldcereal/job.py
VincentVerelst: We can only re-define the metadata manually if we know up front what the metadata (in this case output_labels) will be. Currently we don't know this upfront, as `output_labels` is part of the UDF which is only executed backend-side. 
- For notebooks you could indeed make a client-side script to also determine the output_labels on the client-side
- For UDP's this is unfortunately not possible

kvantricht (approved):
As discussed, confirmed that this all works now we can merge to main (thanks for the big effort!) and think about possible further optimizations later on.

Issue #324, opened 2025-04-02
Job manager migration


jdries:
From GFMap to standard Job Manager

HansVRP:
@kvantricht Which patch extractions which have been ran in the past can we use here as a reference to ensure we generate the same output after the migration?

kvantricht:
> [@kvantricht](https://github.com/kvantricht) Which patch extractions which have been ran in the past can we use here as a reference to ensure we generate the same output after the migration?

what do you need for this? A job ID? a path to where patches are stored? Something else,

kvantricht:
I just took 2 jobs recently finished:

### Sentinel-1 patch extractions

Job ID: `j-2504200912364ba882aada96e107e880`
Reference ID in STAC collection: `2020_LUX_LPIS_POLY_110`
Path to reference output: `/data/worldcereal_data/EXTRACTIONS/SENTINEL_1/2020_LUX_LPIS_POLY_110/31/31UFR`

### Sentinel-2 patch extractions

Job ID: `j-2504181031444e94a63e6e34f33804d1`
Reference ID in STAC collection: `2020_LUX_LPIS_POLY_110`
Path to reference output: `/data/worldcereal_data/EXTRACTIONS/SENTINEL_2/2020_LUX_LPIS_POLY_110/31/31UFR`

### Definition of reference output

For both, the important things to check are:
- correct path construction and file naming
- correct number of output patches extracted (105 S1 ascending, 105 S1 descending, 105 S2)
- Correct file permissions (owned by `wcextractions`, permissions set to `-rwxr-xr-x.`)
- Correct ingestion into STAC collection
- Fully identical file contents (band values, metadata, ..., best to adopt current post-job-action in its entirety)

HansVRP:
@VincentVerelst can you create 2 tickets to add the Sentinel 1 and sentinel 2 patch extraction to the internal APEX framework. 

Below Kristof provides a good list for sanity checks for the job manager migration: 

- [ ] correct path construction and file naming
- [ ] correct number of output patches extracted (105 S1 ascending, 105 S1 descending, 105 S2)
- [ ] Correct file permissions (owned by wcextractions, permissions set to -rwxr-xr-x.)
- [ ] Correct ingestion into STAC collection
- [ ] Fully identical file contents (band values, metadata, ..., best to adopt current post-job-action in its entirety)

These need to become part of the definition of done of the ticket here.

PR #348, opened 2025-05-14
#347-improve-patch-to-point

kvantricht:
#347

### DIFF
--- a/scripts/extractions/patch_to_point.py
+++ b/scripts/extractions/patch_to_point.py
@@ -1,0 +1,454 @@
+import json
+from functools import partial
+from pathlib import Path
+from typing import List, Optional, Union
+
+import geopandas as gpd
+import numpy as np
+import openeo
+import pandas as pd
+import pystac_client
+from loguru import logger
+from openeo import BatchJob
+from openeo.extra.job_management import MultiBackendJobManager
+from openeo_gfmap import Backend, BackendContext, BoundingBoxExtent, TemporalContext
+from openeo_gfmap.utils.catalogue import select_s1_orbitstate_vvvh
+from pandas.core.dtypes.dtypes import CategoricalDtype
+
+from worldcereal.extract.patch_to_point_worldcereal import (
+    create_job_patch_to_point_worldcereal,
+    get_label_points,
+)
+from worldcereal.extract.utils import S2_GRID, upload_geoparquet_artifactory
+from worldcereal.rdm_api.rdm_interaction import RDM_DEFAULT_COLUMNS
+
+
+class EnhancedMultiBackendJobManager(MultiBackendJobManager):
+    def on_job_done(self, job: BatchJob, row):
+        logger.info(f"Job {job.job_id} completed")
+        output_file = generate_output_path_point_worldcereal(self._root_dir, 0, row)
+        job.get_results().download_file(target=output_file, name="timeseries.parquet")
+
+        job_metadata = job.describe()
+        metadata_path = output_file.parent / f"job_{job.job_id}.json"
+        self.ensure_job_dir_exists(job.job_id)
+
+        with metadata_path.open("w", encoding="utf-8") as f:
+            json.dump(job_metadata, f, ensure_ascii=False)
+
+        post_job_action(output_file)
+        logger.success("Job completed")
+
+
+def create_job_dataframe(ref_id, ground_truth_file=None):
+    client = pystac_client.Client.open("https://stac.openeo.vito.be/")
+
+    stac_query = {
+        "ref_id": {"eq": ref_id},
+    }
+
+    search = client.search(
+        collections=["worldcereal_sentinel_2_patch_extractions"],
+        query=stac_query,
+    )
+
+    # Get a list of EPSG codes that occur for this ref_id as we need
+    # to run jobs per UTM zone.
+    logger.info(f"Creating job dataframe for: {ref_id}")
+    logger.info("Looking for unique EPSG codes in STAC collection ...")
+    epsg_codes = {}
+    for item in search.items():
+        epsg = int(item.properties["proj:epsg"])
+
+        if epsg not in epsg_codes and epsg != 4038:
+            logger.debug(f"Found EPSG: {epsg}")
+
+            epsg_codes[epsg] = {
+                "start_date": pd.to_datetime(item.properties["start_date"]),
+                "end_date": pd.to_datetime(item.properties["end_date"]),
+            }
+        elif epsg != 4038:
+            current_start_date = pd.to_datetime(item.properties["start_date"])
+            current_end_date = pd.to_datetime(item.properties["end_date"])
+            if current_start_date > epsg_codes[epsg]["start_date"]:
+                epsg_codes[epsg]["start_date"] = current_start_date
+            if current_end_date < epsg_codes[epsg]["end_date"]:
+                epsg_codes[epsg]["end_date"] = current_end_date
+
+    # Initialize job dataframe for patch to point
+    rows = []
+
+    logger.info(f"Found {len(epsg_codes)} unique EPSG codes in STAC collection.")
+
+    for epsg in epsg_codes.keys():
+        # We assume identical start and end date for the entire ref_id
+        start_date = epsg_codes[epsg]["start_date"]
+        end_date = epsg_codes[epsg]["end_date"]
+
+        # ensure start date is 1st day of month, end date is last day of month
+        # Start a month later and end a month earlier to ensure the extractions cover this.
+        start_date = (start_date + pd.Timedelta(days=31)).replace(day=1)
+        end_date = (
+            end_date.replace(day=1) - pd.Timedelta(days=31) + pd.offsets.MonthEnd(0)
+        )
+
+        # Convert dates to string format
+        start_date, end_date = (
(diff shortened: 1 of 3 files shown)


scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+
+from worldcereal.extract.patch_to_point_worldcereal import (
+    create_job_patch_to_point_worldcereal,
+    get_label_points,
+)
+from worldcereal.extract.utils import S2_GRID, upload_geoparquet_artifactory
+from worldcereal.rdm_api.rdm_interaction import RDM_DEFAULT_COLUMNS
+
+
+class EnhancedMultiBackendJobManager(MultiBackendJobManager):
```
VincentVerelst, line 26: Maybe not for now, but there is an oustanding issue to migrate all job managers: https://github.com/WorldCereal/worldcereal-classification/issues/324

This would lead to multiple subclasses of `MultiBackendJobManager`, maybe even in `src`. So maybe a more clear name could be `PatchToPointJobManager` , to stay futureproof?

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+        gdf["start_date"] = row.start_date
+        gdf["end_date"] = row.end_date
+        gdf["lat"] = gdf.geometry.y
+        gdf["lon"] = gdf.geometry.x
+
+        # Reset index for certain openEO compatibility
+        gdf = gdf.reset_index(drop=True)
+
+        # Upload the geoparquet file to S3
+        logger.info("Deploying geoparquet file to S3 ...")
```
VincentVerelst, line 172: The log info says you're deploying to S3, but you're using the `upload_geoparquet_artifactory` function.

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+        gdfs.append(gdf_filtered)
+
+    # Concatenate after all filtering is done
+    gdf = gpd.GeoDataFrame(pd.concat(gdfs, ignore_index=True))
+
+    # Check for missing attributes based on one columns
+    missing_attrs = gdf["start_date"].isnull()
+    if missing_attrs.sum() > 0.25 * len(gdf):
+        error_msg = (
+            "More than half of the rows have missing attributes. "
```
VincentVerelst, line 315: More than a quarter of the rows?

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+        "ewoc_code": np.int64,
+        "irrigation_status": np.int64,
+        "quality_score_lc": np.int64,
+        "quality_score_ct": np.int64,
+        "extract": np.int64,
+    }
+
+    # Select required attributes and cast to dtypes
+    gdf = gdf[required_attributes.keys()]
+    gdf = gdf.astype(required_attributes)
```
VincentVerelst, line 376: Why not set the dtypes immediately in the `post_job_actions`? This seems out of scope for a function called `merge_individual_parquet_files`

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+
+
+if __name__ == "__main__":
+    root_folder = Path(
+        "/vitodata/worldcereal/tmp/kristof/EXTRACTIONS/WORLDCEREAL/PATCH_TO_POINT/"
+    )
+
+    period = "month"
+    ref_ids = [
+        "2023_FRA_LPIS_POLY_110",
```
VincentVerelst, line 452: Is this something you want hardcoded in the script?

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+        # "2020_FRA_LPIS_POLY_110",
+        # "2021_KEN_COPERNICUS-GEOGLAM-LR_POINT_111"
+    ]
+    restart_failed = True
+
+    logger.info("Starting patch to point extractions ...")
+    logger.info(f"Root folder: {root_folder}")
+    logger.info(f"Period: {period}")
+
+    connection = openeo.connect("openeo-dev.vito.be").authenticate_oidc()
```
VincentVerelst, line 464: Patch-to-point should be able (and is recommended to) to run on the production environment of Terrascope:
`connection = openeo.connect("openeo.vito.be").authenticate_oidc()`

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,474 @@
...
+    merged_gdf.to_parquet(merged_file, index=False)
+    logger.info(f"Merged parquet file saved to: {merged_file}")
+
+    logger.success("ref_id fully done!")
+
+
+if __name__ == "__main__":
+    root_folder = Path(
+        "/vitodata/worldcereal/tmp/kristof/EXTRACTIONS/WORLDCEREAL/PATCH_TO_POINT/"
+    )
```
VincentVerelst, line 448: Is this something you want hardcoded in the script?

VincentVerelst (changes requested):

scripts/extractions/patch_to_point.py
kvantricht: Agreed and fixed.

scripts/extractions/patch_to_point.py
kvantricht: Mmh, latest message with regard to that from Jeroen Dries: "ben bezig om die release er zo snel mogelijk door te krijgen".
Same goes for this weird `jar` path in case you noticed it.

scripts/extractions/patch_to_point.py
kvantricht: and then we need to launch this script each time per `ref_id` and let a wrapper script take care of the looping over multiple ref_ids?

scripts/extractions/patch_to_point.py
VincentVerelst: @kvantricht, there will be a Terrascope prod releas today. Which means we can use the prod backend (openeo.vito.be) and can safely remove the `openeo-jar-path` job option

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,570 @@
...
+        # Get sample points from RDM
+        job_df.loc[ix, "geometry_url"] = url
+
+    # Remove rows without geometry URL as indication for jobs to skip
+    job_df = job_df[job_df["geometry_url"].notna()]
+
+    return job_df
+
+
+def post_job_action(parquet_file):
```
VincentVerelst, line 215: Should we use the same naming convention as for the other types of extractions? E.g. for point extrations, this function is called `post_job_action_point_worldcereal` https://github.com/WorldCereal/worldcereal-classification/blob/main/src/worldcereal/extract/point_worldcereal.py#L180

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,570 @@
...
+        self.ensure_job_dir_exists(job.job_id)
+
+        with metadata_path.open("w", encoding="utf-8") as f:
+            json.dump(job_metadata, f, ensure_ascii=False)
+
+        post_job_action(output_file)
+        logger.success("Job completed")
+
+
+def create_job_dataframe(ref_id, ground_truth_file=None):
```
VincentVerelst, line 44: Same question on naming convention: `create_job_dataframe_patch_to_point_worldcereal`? Same convention as other extraction types: https://github.com/WorldCereal/worldcereal-classification/blob/main/src/worldcereal/extract/point_worldcereal.py#L73

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,570 @@
...
+    }
+
+    # Select required attributes and cast to dtypes
+    gdf = gdf[required_attributes.keys()]
+    gdf = gdf.astype(required_attributes)
+
+    gdf.to_parquet(parquet_file, index=False)
+
+
+def generate_output_path_point_worldcereal(
```
VincentVerelst, line 331: `generate_output_path_patch_to_point_worldcereal`?

scripts/extractions/patch_to_point.py
kvantricht: Good point. I have renamed all methods as suggested but also moved them to the `src/worldcereal/extract/patch_to_point_worldcereal.py` module for consistency with the other extractions where these methods are also defined in `src`.

scripts/extractions/patch_to_point.py
kvantricht: see earlier response

kvantricht:
@VincentVerelst I worked on the suggested changes as described above.

scripts/extractions/patch_to_point.py
```diff
@@ -0,0 +1,242 @@
...
+
+    assert period in ["month", "dekad"], "Period must be either 'month' or 'dekad'."
+
+    # Ref_id output folder
+    output_folder = (
+        root_folder / ref_id if period == "month" else root_folder / f"{ref_id}_10D"
+    )
+    output_folder.mkdir(parents=True, exist_ok=True)
+
+    job_tracking_csv = output_folder / "job_tracking.csv"
```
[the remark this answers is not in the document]
kvantricht: Thanks, this is now updated in the code.

PR #351, opened 2025-05-20
Remove combined use of save_result and out_format in execute_batch

### LINKED ISSUE #350: Fix bug in inference process graph
As reported by users:

`OpenEoClientException: DataCube.execute_batch() with explicit output format 'GTiff', but the process graph already has `save_result` node(s) which is ambiguous and should not be combined.`

The currently used openEO client version raises an exception on the inference process graph due to a combination of `save_result` and `execute_batch(format=...)` which is no longer accepted.

We should fix the inference pipeline by making the final `save_result` explicit and no longer depend on the `format` option of `execute_batch`.

kvantricht:
Closes #350.

### DIFF
--- a/src/worldcereal/job.py
+++ b/src/worldcereal/job.py
@@ -191,7 +191,7 @@ def generate_map(
         )
 
     # Save the final result
-    classes.save_result(
+    classes = classes.save_result(
         format=out_format,
         options=dict(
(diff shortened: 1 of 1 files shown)


jdegerickx (approved):
looks good.
Hopefully the tests succeed...

PR #360, opened 2025-05-26
new script demonstrating the use of croptype udp

jdegerickx:
script that can be used by gisat to start integrating crop type udp in processing hub...


scripts/inference/croptype_mapping_udp.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Temporal extent needs to consist of exactly twelve months,
+# always starting first day of the month and ending last day of the month.
+temporal_extent = ["2018-11-01", "2019-10-30"]
+
+# Provide the link to your custom model
+model_url = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/models/PhaseII/downstream/tests/be_multiclass-test_custommodel.onnx"
+
+# Optional parameters
+orbit_state = "ASCENDING"  # optional parameter, default is "DESCENDING"
```
kvantricht, line 36: We should maybe state explicitly that this should actually never be kept default because in many regions this results in S1 being unavailable?

scripts/inference/croptype_mapping_udp.py
```diff
@@ -0,0 +1,60 @@
...
+postprocess_method = "smooth_probabilities"  # optional parameter, available choices are "majority_vote" and "smooth_probabilities". Default is "smooth_probabilities"
```
kvantricht, line 37: I thought at some point we switched to majority voting because it was faster and giving better results? If it's not the default we may still want to add it here for Gisat to use those settings?

scripts/inference/croptype_mapping_udp.py
```diff
@@ -0,0 +1,72 @@
...
+}
+
+# Temporal extent needs to consist of exactly twelve months,
+# always starting first day of the month and ending last day of the month.
+# Ideally, the season of interest should be nicely centered within the selected twelve months period.
+temporal_extent = ["2018-11-01", "2019-10-30"]
+
+# Provide the link to your custom model
+# The model should be in ONNX format and publicly accessible.
+model_url = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/models/PhaseII/downstream/tests/be_multiclass-test_custommodel.onnx"
```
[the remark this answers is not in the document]
jdegerickx: would not do this, as we currently don't have a default model that we would like to promote. That is why model_url is listed here as an essential parameter, not an optional one.

jdegerickx:
> Looks good; should we document what is the expected result? They cannot specify whether or not intermediate results should be saved? Is the default yes or no?

Will add some documentation indeed, just like I did for the cropland udp script

kvantricht:
> > Looks good; should we document what is the expected result? They cannot specify whether or not intermediate results should be saved? Is the default yes or no?
> 
> Will add some documentation indeed, just like I did for the cropland udp script

ah okay currently we don't have crop type in processing module yet? In this case we need to document how they can train their own model?

jdegerickx:
> > > Looks good; should we document what is the expected result? They cannot specify whether or not intermediate results should be saved? Is the default yes or no?
> > 
> > 
> > Will add some documentation indeed, just like I did for the cropland udp script
> 
> ah okay currently we don't have crop type in processing module yet? In this case we need to document how they can train their own model?

agreed, will add some explanation and link to the crop type notebook

PR #362, opened 2025-06-02
Run inference workflow on Python 3.11 image of openeo

VincentVerelst:
- Updated gfmap dependency version to 0.4.6
- `generate_map` now executes a batch job on the python 3.11 image of openeo


kvantricht (approved):
Looks good to me if the tests on the backend were successful.

PR #366, opened 2025-06-18
map ui also supports shp and gpkg upload

jdegerickx:
supported file formats = .gpkg and .zip (the latter represents a zipped shapefile)

different cases are supported:
- file contains one polygon --> polygon is converted to its bbox
- file contains multiple polygons --> we compute the total bounds
- file contains mulitple points --> we compute total bounds
- file contains single point --> not allowed because impossible to generate an extent without specifying a buffer...

### DIFF
--- a/src/worldcereal/utils/map.py
+++ b/src/worldcereal/utils/map.py
@@ -1,4 +1,8 @@
+import json
+import tempfile
 import warnings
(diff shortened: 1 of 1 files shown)


kvantricht (approved):
Nice work!

PR #367, opened 2025-06-18
Added large-scale inference handler

kvantricht:


jdegerickx:
@kvantricht --> as you created the branch, I couldn't make you official reviewer, but can you please go through and check?

fully revised version of custom crop type mapping notebook, now with support for large area mapping.

Needed to break down run_large_inference function to be able to add custom notebook stuff, but all things related to notebooks remain in notebook utils.
Combined the function with the utils created by Hans to easily monitor production.
If you want to test, you can draw a small area and set tiling_resolution to e.g. 5 --> will create 5x5 km tiles.

Visualization of the final result is simplified and deleted the leafmap option as that one will never work anyways.

On the side also updated the date_slider and added functionality for user to discard some classes before training a model.

src/worldcereal/job.py
```diff
@@ -1,10 +1,31 @@
-"""Executing inference jobs on the OpenEO backend."""
+"""Executing inference jobs on the OpenEO backend.
```
kvantricht, line 1: @jdegerickx can you update this top-level docstring to describe the purpose of the new `prepare_largescale_inference` entry point?

src/worldcereal/job.py
```diff
@@ -201,6 +320,99 @@ def create_inference_process_graph(
...
+    ----------
+    row : pd.Series
+        _description_
+        Contains at least the following fields:
+        - start_date: str, start date of the temporal extent
+        - end_date: str, end date of the temporal extent
+        - geometry: shapely.geometry, geometry of the spatial extent
+        - tile_name: str, name of the tile
+        - epsg: int, EPSG code of the spatial extent
+        - bounds_tiling_epsg: str, bounds of the spatial extent in EPSG coordinates
```
kvantricht, line 348: required format? I saw you worked with "xx;yy;zz;aa". In the past we used to work with true bounds tuple "(xx, yy, zz, aa" which can be loaded with something like `eval(bounds_tuple)`. More elegant than the `split(";")` method and also a bit shorter in the creation step as well. Is this something that could still be considered?

src/worldcereal/job.py
```diff
@@ -406,3 +607,219 @@ def collect_inputs(
...
+    This function orchestrates the execution of large-scale inference jobs
+    using a production grid (either a Parquet file or a GeoDataFrame) and specified parameters.
+    It manages job creation, tracking, and execution on the OpenEO backend.
+
+    Parameters
+    ----------
+    production_grid : Union[Path, gpd.GeoDataFrame]
+        Path to the production grid file in Parquet format or a GeoDataFrame.
+        The grid must contain the required attributes: 'start_date', 'end_date',
+        'geometry' (in EPSG:4326), 'tile_name', 'epsg' and 'bounds_tiling_epsg'.
```
kvantricht, line 636: why is 4326 compulsory for the dataframe?

src/worldcereal/job.py
```diff
@@ -406,3 +607,219 @@ def collect_inputs(
...
+    job_manager.run_jobs(
+        df=job_df,
+        start_job=start_job,
+        job_db=job_tracking_csv,
+    )
+
+    logger.info("Job manager finished.")
+
+
+def prepare_largescale_inference(
```
kvantricht, line 693: in case we're resuming jobs, this is not really "preparation". Wondering if we can find a more accurate name for this function?

src/worldcereal/job.py
```diff
@@ -406,3 +607,219 @@ def collect_inputs(
...
+        output_dir = Path(output_dir)
+        output_dir.mkdir(parents=True, exist_ok=True)
+
+    # Make a connection to the OpenEO backend
+    connection = BACKEND_CONNECTIONS[backend_context.backend]()
+
+    # Setup the job manager
+    logger.info("Setting up the job manager.")
+    manager = InferenceJobManager(root_dir=output_dir)
+    manager.add_backend("cdse", connection=connection, parallel_jobs=parallel_jobs)
```
kvantricht, line 766: this may have been overlook from my side but we should not hardcode `cdse` here and take the value from `backend_context.backend` (if that is the right syntax).

kvantricht (commented):
Somme comments added, now will try the notebook.

src/worldcereal/job.py
jdegerickx: If you're resuming jobs, you would still need to run the "run_largescale_inference" function. This preparation function is just called in that run function. 
So I would be inclined to keep the name?

src/worldcereal/job.py
jdegerickx: I agree, will try to rework that!

src/worldcereal/job.py
kvantricht: isn't it rather a `setup_job_manager` function or something like that?

PR #374, opened 2025-07-03
Fix tz awareness point extractions

kvantricht:
This PR contains a fix so (private) point extractions are now also tz-naive. Furthermore, these extractions will now also contain `S2-L2A-B8A` and some additional attribute checking is implemented.

### DIFF
--- a/src/worldcereal/extract/patch_to_point_worldcereal.py
+++ b/src/worldcereal/extract/patch_to_point_worldcereal.py
@@ -1,8 +1,8 @@
+import copy
 from pathlib import Path
 from typing import Optional, Union
 
 import geopandas as gpd
-import numpy as np
 import openeo
 import pandas as pd
 import pystac_client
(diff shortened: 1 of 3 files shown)


kvantricht:
> Maybe risky to define the required attributes and associated data types in point_worldcereal.py? Might not be fully in sync with patch extractions or attribute check that we do during ref data handling?

i think the one during ref data handling is quite different because there we don't check datatypes from extractions; but good point in the duplication of the dictionary in both patch-to-point as in point-worldcereal, which should be the same. I will centralize these two.

jdegerickx:
FYI, this is the attribute check I do when harmonizing reference data, but it's in a different repo: 
https://github.com/WorldCereal/worldcereal-referencedata/blob/1f656f5a562dc8e198ca140d305978cd464d58b8/harmonization/harmonize_script_vito.py#L26

jdegerickx (approved):
Looks good, can't wait to try :)

PR #384, opened 2025-07-16
Check STAC for existing extractions

cbutsko:
Here's a basic check in STAC catalog for existing samples. 
**Important things to consider:**
- this check will **only happen if no `job_tracking.csv` exists**; it might be interesting to extend it to the case when `job_tracking.csv` exists.
- it does slow things down, but not too much. For **a big dataset with ~10K samples** marked for extraction the search takes **~20s** 
- **by default it's `False`**, not to intervene with the existing logic. Maybe once we are comfortable with functionality, we should change the default behavior to `True`
- no tests are added, since I couldn't think of any proper tests ☹ though I did test it locally with existing datasets, and it works as expected, filtering out samples that are found in relevant stac collection

### DIFF
--- a/src/worldcereal/extract/common.py
+++ b/src/worldcereal/extract/common.py
@@ -330,11 +330,9 @@ def load_dataframe(
 
             stac_search = client.search(
                 collections=[STAC_COLLECTION],
-                filter={"op": "=", "args": [{"property": "properties.ref_id"}, ref_id]}, 
+                filter={"op": "=", "args": [{"property": "properties.ref_id"}, ref_id]},
                 filter_lang="cql2-json",
-                fields={
-                    "exclude": ["assets", "links", "geometry", "bbox"]
-                },
+                fields={"exclude": ["assets", "links", "geometry", "bbox"]},
             )
             for item in stac_search.items():
                 sample_id = item.properties.get("sample_id")


kvantricht (commented):
Nice! Looks quite good to me. Can you also request a review from @VincentVerelst ?

src/worldcereal/extract/common.py
```diff
@@ -288,9 +300,57 @@ def load_dataframe(df_path: Path, extract_value: int = 0) -> gpd.GeoDataFrame:
...
+                )
+            else:
+                # should it be softer somehow and just continue to the next dataset if available?..
+                raise Exception(
+                    f"All samples already exist in STAC API for ref_id {ref_id}, collection {collection}. No samples to extract. Exiting."
+                )
+        else:
+            pipeline_log.warning(
+                "STAC check is only performed for PATCH_SENTINEL1 or PATCH_SENTINEL1 collections. ",
+                "Collection %s is not supported. Skipping STAC check.",
```
kvantricht, line 350: this message will be confusing if collection = None ( the default)

src/worldcereal/extract/common.py
```diff
@@ -288,9 +300,57 @@ def load_dataframe(df_path: Path, extract_value: int = 0) -> gpd.GeoDataFrame:
...
     if df_path.name.endswith("parquet"):
-        return gpd.read_parquet(df_path, filters=filters)
+        df = gpd.read_parquet(df_path, filters=filters)
     else:
-        return gpd.read_file(df_path, filters=filters)
+        df = gpd.read_file(df_path, filters=filters)
+
+    if check_existing:
+        if collection in ["PATCH_SENTINEL1", "PATCH_SENTINEL2"]:
+            ref_id = str(df_path).split("/")[-1].split(".")[0]
```
kvantricht, line 309: the method that calls `load_dataframe` has already access to `ref_id`, so I would pass it as argument to `load_dataframe` instead of inferring it from the filename, which potentially in the future may be problematic.

src/worldcereal/extract/common.py
```diff
@@ -288,9 +301,62 @@ def load_dataframe(df_path: Path, extract_value: int = 0) -> gpd.GeoDataFrame:
...
+            else:
+                raise ValueError(
+                    f"Collection {collection} is not supported for STAC check."
+                )
+
+            stac_search = client.search(
+                collections=[STAC_COLLECTION], query=stac_query, max_items=None
+            )
+            for item in stac_search.items():
+                sample_id = item.properties.get("sample_id")
```
VincentVerelst, line 336: I guess we need to get the actual sample_id from the item properties and can't just use the `item.id`? I remember some ID's have the `openEO_` prefix and some don't. Correct?

src/worldcereal/extract/common.py
```diff
@@ -288,9 +301,62 @@ def load_dataframe(df_path: Path, extract_value: int = 0) -> gpd.GeoDataFrame:
...
+                STAC_COLLECTION = "worldcereal_sentinel_1_patch_extractions"
+            elif collection == "PATCH_SENTINEL2":
+                STAC_COLLECTION = "worldcereal_sentinel_2_patch_extractions"
+            else:
+                raise ValueError(
+                    f"Collection {collection} is not supported for STAC check."
+                )
+
+            stac_search = client.search(
+                collections=[STAC_COLLECTION], query=stac_query, max_items=None
```
VincentVerelst, line 333: I recently discovered our STAC API supports CQL2 filters. This is faster than the query parameter. Below a code snippet on how to use it:

```python
stac_search = client.search(
    collections=[STAC_COLLECTION],
    filter={"op": "=", "args": [{"property": "properties.ref_id"}, ref_id]}, 
    filter_lang="cql2-json",
    fields={
        "exclude": ["assets", "links", "geometry", "bbox"]
    },
)
```
This query also excluded things like asset metadata, since we don't need it here. 

I benchmarked the query-approach vs the filter-approach on the `2022_EU_LUCAS_POLY_111` dataset:

- query-approach: 3m30s
- filter-approach: 2m45s

So this might speed up things slightly

cbutsko: nice! replaced my query with your search here 31df596d6c636ee34585a87b764be162eae2ce8c
also tested it on a couple datasets - indeed, makes things a little faster 👍

src/worldcereal/extract/common.py
kvantricht: @VincentVerelst we are stripping those things away but somewhow we are never resetting the item ID, so this is why since quite some time the stac items have this pattern. We probably should not change this now.

kvantricht (approved):

PR #421, opened 2025-09-10
Add the option to have controlled val and test splits

kvantricht:


src/worldcereal/train/data.py
```diff
@@ -199,14 +200,14 @@ def get_training_dfs_from_parquet(
     # Remove classes with too few samples for stratification
     df = remove_small_classes(df, min_samples=10)
 
-    if val_samples_file is not None:
-        logger.info(f"Controlled train/test split based on: {val_samples_file}")
-        val_samples_df = pd.read_csv(val_samples_file)
+    if test_samples_file is not None:
+        logger.info(f"Controlled train/test split based on: {test_samples_file}")
```
giollimirgia, line 204: perhaps "trainval/test split" in the logger info since we are gonna split train into train and val anyway afterwards

src/worldcereal/train/data.py
```diff
@@ -216,16 +217,24 @@ def get_training_dfs_from_parquet(
...
-        # With controlled validation set it's possible that either
-        # the validation set has unique classes not present in training
+    if val_samples_file is not None:
+        logger.info(f"Controlled train vs val split based on: {val_samples_file}")
+        val_samples_df = pd.read_csv(val_samples_file)
+        train_df, val_df = split_df(
+            trainval_df, val_sample_ids=val_samples_df.sample_id.tolist()
+        )
+    else:
+        train_df, val_df = train_test_split(
```
giollimirgia, line 228: unless this function is also used somewhere else and the test is actually meant to be the independent set for final performance evaluation, i would change the name `train_test_split` to `train_val_split` to be aligned with the definition we are using.

kvantricht: you mean this one?

`from sklearn.model_selection import train_test_split`

wouldn't rename official external modules and stick to the official name instead.

giollimirgia: ah ok didn't see it was from sklearn. then it's fine

giollimirgia (approved):

PR #429, opened 2025-09-15
World cereal description minor change

tho-hole:
I've checked the text with Jeroen Degerickx and this version is approved.


jdegerickx (approved):
OK!

PR #430, opened 2025-09-15
process_parquet bugs

cbutsko:
This PR addresses the following:

- more verbose logging of the inner workings of `process_parquet`
- bug fixes:
    - incorrect handling of range that resulted in padding with wrong number of dummy timesteps (used to be one less than needed) and that failed if `min_edge_buffer = 1`
    - allowing `min_edge_buffer` to be 0
    - removing parameter `required_min_timesteps` so that now it's inferred from frequency and the check is mandatory
- adding a few more tests


src/worldcereal/utils/refdata.py
```diff
@@ -691,7 +691,7 @@ def process_extractions_df(
         )
 
     df_processed = process_parquet(
```
jdegerickx, line 693: in this process_extractions_df function we have "valid_time_buffer" as a parameter. Should this be aligned with "min_edge_buffer" in process_parquet? So should we pass this valid_time_buffer explicitly here? Or can the two types of buffers be different from one another?

PR #455, opened 2025-10-15
Combined udf

HansVRP:


HansVRP:
> @HansVRP I suggest this configuration for starters:
> 
> ```
> bbox = [3.84834727, 49.76688828, 4.14813588, 49.96072086]
> start_date = "2023-01-01"
> end_date = "2023-12-31"
> ```
> 
> Spatially, it's a block of HRL LAEA grid (20x20km) in the north of France (`E388N298`), close to Belgium. I'll use the same one for my one tests, and will in parallel think of other patches in different regions. If you need in in UTM, here on Terrascope is a subsetted gpkg with only this polygon: `/vitodata/worldcereal/tmp/cbutsko/inference_patches/10D/laea_grid_subset.gpkg`
> 
> Additionally, I have added a file with three 10x10km patches in BRA, THA and GRC. There, the file also contains start and end dates for each patch. Here's the path: `/vitodata/worldcereal/tmp/cbutsko/inference_patches/worldcereal_test_patches_subset.gpkg`
> 
> Let me know if that's sufficient for you!

could you provide me the coordinates in the utm projection? I currently cannot access the worldcereal storage

HansVRP:
@kvantricht I added a unit test to provide some clarity on how I was running the dual_inference UDF. 
ruff will btw always file due to how we need to pull in onnxruntime inside the udf

HansVRP:
@kvantricht unit test are passing now and I also managed to run it in openEO. So feel free to take a look.  What would be good is to create a follow-up ticket on how this UDF should be integrated in the worldcereal repo. I am not knowledgeable on the set-up here.

I am also assigning @jdries as reviewer to provide feedback on the model caching; getting the presto embeddings, threading and whatever catches his eye

@cbutsko if you could notify me of the UTM validation bboxes; I can continue experimental validation

PR #457, opened 2025-10-16
Notebook updates

jdegerickx:
- new notebook for the training next week with a use case where participants train a model for france and apply to czech republic
- visualization of the final result  (crop type map) is now again possible with leafmap. Default way of visualization is still matplotlib...

---> requires new build of image with updated dependencies!

- when visualizing time series of extractions, people now also see the crop type label of each sample

### DIFF
--- a/notebooks/notebook_utils/extractions.py
+++ b/notebooks/notebook_utils/extractions.py
@@ -493,9 +493,18 @@ def query_extractions(
     return merged_df
 
 
-def retrieve_extractions_extent() -> gpd.GeoDataFrame:
+def retrieve_extractions_extent(include_crop_types: bool = True) -> gpd.GeoDataFrame:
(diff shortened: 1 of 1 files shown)


notebooks/notebook_utils/extractions.py
```diff
@@ -468,8 +477,35 @@ def query_extractions(
...
+    print(f"Represented crop groups: {merged_df['sampling_label'].unique()}")
+
     # Explictily drop column "feature_index" if it exists
     if "feature_index" in merged_df.columns:
         merged_df = merged_df.drop(columns=["feature_index"])
 
     return merged_df
+
+
+def retrieve_extractions_extent() -> gpd.GeoDataFrame:
```
kvantricht: we could consider filtering yes/no on croptype datasets vs landcover datasets?

kvantricht (approved):

PR #498, opened 2025-11-17
Patch extractions updates

cbutsko:
An update that allows more efficient reading of heavy input datasets using duckdb.

### DIFF
--- a/src/worldcereal/extract/common.py
+++ b/src/worldcereal/extract/common.py
@@ -84,12 +84,7 @@ def extraction_job_quality_check(
     Exception
         Raised if the SAR backscatter error ratio exceeds the threshold.
     """
-    # TEMPORARY FIX FOR MISMATCH IN STAGING NAME; OTHER BACKENDS ARE NOT AFFECTED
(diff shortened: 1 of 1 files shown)


src/worldcereal/extract/common.py
```diff
@@ -83,7 +84,12 @@ def extraction_job_quality_check(
     Exception
         Raised if the SAR backscatter error ratio exceeds the threshold.
     """
-    conn = BACKEND_CONNECTIONS[Backend[job_entry["backend_name"].upper()]]()
+    # TEMPORARY FIX FOR MISMATCH IN STAGING NAME; OTHER BACKENDS ARE NOT AFFECTED
```
kvantricht: do we want this temporary fix to be present in main?

cbutsko: no, indeed. I checked that the staging name mismatch was happening in the local version of openeo-gfmap (which accommodates for some private backends), so I just changed it there. and then nothing needs to be modified at worldcereal side, so I reverted the change fcd45240617d3e60b2c3723886c85cae36da385c. thanks!

PR #500, opened 2025-11-19
Updating prometheo wheel version in inference

cbutsko:
For some reason, inference.py had older version of prometheo wheel. Need to update.


VincentVerelst:
@kvantricht, noted

PR #502, opened 2025-11-20
Update croptype udp

VincentVerelst:
- Inference UDPs now contain latests changes in the inference workflow
- Two notebooks have been added and/or modified that I used to create the UDPs, would be good to also have these in the main branch for the future

In case the PR is approved, I'll tag the new UDPs to give them a final namespace. I propose to rename them both to v2.0.0, as there are some major/breaking changes introduced.


kvantricht (approved):
looks good!

PR #503, opened 2025-11-25
Fix private extractions ntb

jdegerickx:
Based on experiences from FAO workshop, two fixes were done to ensure smooth execution of the private extractions notebook.
1. Ensure the visualize_timeseries function works if less than 5 samples available in the dataframe
2. Avoid errors when users do not read instructions and accidentally execute an optional block of code


kvantricht:
i am retriggering the tests to see if the unrelated failure was a temporary glitch

PR #523, opened 2025-12-11
Default value of extract_value should be 0

kvantricht:
It's safer to by default extract all samples and just make sure that our internal patch extractions set `extract_value=1`, which is already the case. At the moment, also private extractions were working correctly (setting `extract_value=0`), but it's more future proof like this.

### DIFF
--- a/src/worldcereal/extract/common.py
+++ b/src/worldcereal/extract/common.py
@@ -893,7 +893,7 @@ def run_extractions(
     job_options: Optional[Dict[str, Union[str, int]]] = None,
     parallel_jobs: int = 2,
     restart_failed: bool = False,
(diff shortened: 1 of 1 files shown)


jdegerickx (commented):
does it make sense to change the default value only here, while the default in function run_extractions remains 1?

kvantricht:
> does it make sense to change the default value only here, while the default in function run_extractions remains 1?

excellent catch and an overview from my side! Fixed in new commit.

jdegerickx:
> > does it make sense to change the default value only here, while the default in function run_extractions remains 1?
> 
> excellent catch and an overview from my side! Fixed in new commit.

I was also referring to the run_extractions function in common.py...

PR #537, opened 2026-01-09
Add CRS to queried extractions

### LINKED ISSUE #489: Queried extractions should have a CRS
Currently, geopandas GeoDataFrame resulting from querying extractions has no defined CRS, which hampers visualization in e.g. QGIS. We should set the CRS.

kvantricht:
Closes #489. Queried extractions now result in `geopandas.GeoDataFrame` with valid CRS.


jdegerickx (approved):
looks fine to me!

PR #555, opened 2026-01-29
324 migrate job manager

VincentVerelst:
Made minimal changes to the codebase in order to get rid of the openeo<=0.35.0 dependency. 

The `run_extractions` function now doesn't use the GFMap Job Manager anymore, but its usage should remain unchanged.

I've already tested it locally and it seems to work. @kvantricht , I don't know who is currently running patch extractions for WorldCereal, but it would be great if they could start using this branch for the extractions for some additional testing, before we merge to main.

### DIFF
--- a/tests/worldcerealtests/testresources/preprocess_graph.json
+++ b/tests/worldcerealtests/testresources/preprocess_graph.json
@@ -580,6 +580,7 @@
                 "DEM"
             ],
             "id": "COPERNICUS_30",
+            "properties": {},
             "spatial_extent": null,
             "temporal_extent": null
         }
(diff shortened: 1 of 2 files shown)


kvantricht:
> > the tests fail on the process graph
> 
> I still need to look into that one. Is that failing test specific for this PR? Because I also see failing tests on main

the ones that failed on main were an issue with the extraction bucket on s3. We fixed that (metadata of parquet issue) so those tests work again. The ones that fail in this PR are processing graphs so they are specific to this PR indeed.

cbutsko (approved):
To me, the proof is in the pudding: patch extractions ran successfully with this branch, followed by patch-to-point, and all data is looking valid. Good for me to merge!

PR #569, opened 2026-02-11
More local inference updates

cbutsko:
Mainly adding better local tiff writing and reserving this for other minor changes to notebook that might come along the way.


jdegerickx:
these changes have been taken on board in another PR...

PR #664, opened 2026-04-10
Updates to user-facing downstream training workflow

### LINKED ISSUE #627: Verify quality and outlier score handling in user-facing downstream model training
We have to make sure users are not using samples with quality scores of 0, and ideally should weight the samples using their quality score.

kvantricht:
### PRIOR ANALYSIS

`align_to_season`:

- "_Drops irrelevant samples ~~and aligns your data to the selected season~~_": the second part is no longer true
- Processing period is not passed to `process_parquet`
- Probably better results could be expected if alignment would happen, but many extractions will be dropped because they do not fit the specified 12 months (due to the 18 month extraction window)
- valid_time_buffer=MIN_EDGE_BUFFER which is currently 2; problematic when user collects data today and wants to train a model tomorrow (valid_time at the edge) -> we should support this case.

`min_season_coverage`
- Was not exposed yet, defaults to `1`
- Many samples silently behave weirdly (partial overlap means no mask activation which then triggers global pooling!) 

`mask_on_training` comes with its own default masking config;
- We should check and if needed update the parameters.

`augment`
- Set to `False` and not exposed. Is this what we want? I am not sure.

`test_size`
- Proposed Test size in app is 0.2, in function normally 0.15. Is this intended?

Quality scores, outliers and sample weighting
- User-facing workflow does not use any of this yet.

### CHANGES

`align_to_season`
- `valid_time_buffer` now defaults to 0. Tests did not show errors so far. 

`min_season_coverage`
- Parameter exposed in user-facing workflow; not yet in app
- By default now set to 0.5 (requiring at least half coverage)
- Violation of a sample now leads to it being dropped and a warning is logged, explaining the cause and suggesting fixes

Scoring mechanism
- Extractions in public bucket now contain the outlier information
- Quality and outlier scoring is now used to (i) drop and (ii) weight samples

Minor
- A `RuntimeError` is raised when fewer than the minimum number of crop types remain after filtering (instead of silently continuing).
- Deprecated CatBoost functionality was removed

### TODO

- [x] Update description of `align_to_season` in app (@jdegerickx)
- [x] Update parameters of masking config (@cbutsko you have a suggestion?)
- [x] Review proposed `test_size` in app (@jdegerickx )
- [x] Add `min_season_coverage` parameter to the app (@jdegerickx )


cbutsko:
My best suggestion regarding masking config would be to set the defaults the same as we used during finetuning. For that, almost all parameters are already the same, with the two exceptions:
* propose to set `s1_timestep_dropout_prob` to 0.05 instead of 0.1
* set `meteo_timestep_dropout_prob` to 0.05 instead of 0.03
I think both changes will have very minor effect, tbh. 
What can maybe be more beneficial is possibility to disable S1 completely, for example? And thus expose this as a separate argument? In case it's a very localised run in an area with limited coverage. Not sure, maybe it's rather "nice-to-have" thing.

notebooks/notebook_utils/classifier.py
```diff
@@ -88,7 +78,7 @@ def align_extractions_to_season(
     df: pd.DataFrame,
     season: Optional[TemporalContext] = None,
```
jdegerickx: shouldn't we call this `processing_period` to avoid confusion with `season_window` and be aligned with downstream functions?

jdegerickx:
> test_size
> Proposed Test size in app is 0.2, in function normally 0.15. Is this intended?

No particular reason --> I believe I just implemented the value that was used at some point in one of the example scripts we had...

jdegerickx:
> augment
> Set to False and not exposed. Is this what we want? I am not sure.

@kvantricht I recall we had a conversation about this a while ago and you specifically recommended to set it this way. (because originally it was exposed to the user)
What would be pro's and con's of exposing in your opinion?

jdegerickx:
> min_season_coverage
> Was not exposed yet, defaults to 1
> Many samples silently behave weirdly (partial overlap means no mask activation which then triggers global pooling!)

@kvantricht, I can expose this parameter in the app, no problem, but how should the accompanying explanation look like? 
We need to provide:
- a clear definition of this parameter (the description in the function signature is not 100% clear to me)
- a guideline to the users --> in which specific cases should they change the default of 0.5? Which value should they pick?
Any ideas here?

kvantricht:
> My best suggestion regarding masking config would be to set the defaults the same as we used during finetuning. For that, almost all parameters are already the same, with the two exceptions:
> 
> * propose to set `s1_timestep_dropout_prob` to 0.05 instead of 0.1
> * set `meteo_timestep_dropout_prob` to 0.05 instead of 0.03
>   I think both changes will have very minor effect, tbh.
>   What can maybe be more beneficial is possibility to disable S1 completely, for example? And thus expose this as a separate argument? In case it's a very localised run in an area with limited coverage. Not sure, maybe it's rather "nice-to-have" thing.

are these the updated values after a discussion we recently had to increase the values? Good point about S1 but the problem is that currently we cannot control it in inference and user likely will not be able any time soon. If the location has hardly any S1, the training data shouldn't have it either and we train for it. The problem occurs when training data is from a time with S1 and inference without, but with S1C now operational maybe it's less common. We should probably not give this power to the user just during training;

notebooks/notebook_utils/classifier.py
kvantricht: Good point, updated.

jdegerickx (approved):

PR #667, opened 2026-04-13
Landcover-only dataset fixes

kvantricht:
- Handling of quality score for LC-only datasets is now fixed: quality score of LC used in CT mapping in case of LC-only dataset
- Seasonal mask handling for LC-only datasets is now fixed: shortcut to all-valid seasonal masks in case of LC-only dataset


jdegerickx (approved):
looks good to me!

PR #669, opened 2026-04-14
Several fixes for outlier (score) handling

kvantricht:
- Centralize `OUTLIER_COLUMNS` so we no longer have multiple places where column names are hardcoded
- Outliers are no longer dropped from the dataframe, but instead their weights are put to 0; this way we can have diverging outlier handling for landcover vs croptype.
- Evaluation metrics only take into account samples that have a task-specific weight above a certain threshold (default=0)

### DIFF
--- a/notebooks/notebook_utils/classifier.py
+++ b/notebooks/notebook_utils/classifier.py
@@ -380,7 +380,7 @@ def train_seasonal_torch_head(
         quality_col=quality_col,
         outlier_score_col=outlier_score_col,
         outlier_col=outlier_col,
-        outlier_drop_level="drop_candidate",
+        outlier_drop_mode="drop_candidate",
         zero_quality_cols=zero_quality_cols,
         **trainer_kwargs,
     )
--- a/notebooks/notebook_utils/classifier.py
+++ b/notebooks/notebook_utils/classifier.py
@@ -22,6 +22,7 @@ from openeo_gfmap import TemporalContext
 from prometheo.utils import DEFAULT_SEED
 from tabulate import tabulate
 
+from worldcereal.train import OUTLIER_COLUMNS
 from worldcereal.train.backbone import (
     build_presto_backbone,
     checkpoint_fingerprint,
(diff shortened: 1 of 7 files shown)


cbutsko (approved):
Looks good! I read through the tests, seems like they are covering all handling of **_a_** outlier column, and maybe this is enough at this point.
I was also thinking it might be useful to add a few tests to cover interaction cases between LC/CT outlier flags and scores, but that might be a bit excessive here.

kvantricht:
> Looks good! I read through the tests, seems like they are covering all handling of **_a_** outlier column, and maybe this is enough at this point. I was also thinking it might be useful to add a few tests to cover interaction cases between LC/CT outlier flags and scores, but that might be a bit excessive here.

I see what you mean, though `attach_sample_weights` does act on only one column so its functionality does not deal with this kind of interaction, right?

uceshaf:
For
https://github.com/WorldCereal/worldcereal-classification/blob/3bd56ae1d2041ee1cfa7ecedf3ae595ae335e05b/src/worldcereal/train/finetuning_utils.py#L238-L249
We should first check if drop_level == "keep" before giving any warnings
```
if drop_level == "keep":
    logger.info(...)
    return pd.Series(False, index=df.index, dtype=bool)

if not outlier_col or outlier_col not in df.columns:
    if outlier_col:  # only warn if a column was actually expected
        logger.warning(...)
    return pd.Series(False, index=df.index, dtype=bool)
```

uceshaf:
Very minor: 
https://github.com/WorldCereal/worldcereal-classification/blob/3bd56ae1d2041ee1cfa7ecedf3ae595ae335e05b/src/worldcereal/train/finetuning_utils.py#L100
is
`**outlier_score** – read from *outlier_score_col*; a continuous confidence `

kvantricht:
> For
> 
> https://github.com/WorldCereal/worldcereal-classification/blob/3bd56ae1d2041ee1cfa7ecedf3ae595ae335e05b/src/worldcereal/train/finetuning_utils.py#L238-L249
> 
> 
> We should first check if drop_level == "keep" before giving any warnings
> ```
> if drop_level == "keep":
>     logger.info(...)
>     return pd.Series(False, index=df.index, dtype=bool)
> 
> if not outlier_col or outlier_col not in df.columns:
>     if outlier_col:  # only warn if a column was actually expected
>         logger.warning(...)
>     return pd.Series(False, index=df.index, dtype=bool)
> ```

good point, fixed!

kvantricht:
> Very minor:
> 
> https://github.com/WorldCereal/worldcereal-classification/blob/3bd56ae1d2041ee1cfa7ecedf3ae595ae335e05b/src/worldcereal/train/finetuning_utils.py#L100
> 
> 
> is
> `**outlier_score** – read from *outlier_score_col*; a continuous confidence `

also fixed.

uceshaf (approved):

PR #672, opened 2026-04-15
Data read optimizations

uceshaf:
### Summary
Reduced memory footprint and sped up the preprocessing pipeline, enabling all experiments to run within **40 GB memory** (previously OOM).

---

#### 1. Parquet Row Group Preservation
The original wide parquet had **435 row groups** (~20k rows each), allowing efficient chunked reads. However, every time the file was rewritten after region assignment, it collapsed into **a single row group**, and caused slow sequential reads and OOM on subsequent runs.

**Fix:** Save-back now uses `pyarrow.parquet.write_table` with `row_group_size=20_000` and sets `g+w` permissions right away.

#### 2. Eliminated Redundant In-Memory Copies During Region Calcs
The region assignment path previously created multiple copies of the full ~7M-row DataFrame. Refactored to mutate in-place and free intermediate GeoDataFrames asap

#### 3. Spatial Join for Region Assignment
Region assignment used `sjoin_nearest` on all ~7M points which was a big bottleneck that took **>30 minutes**.

**Fix:** Two-pass option:
1. **`sjoin(..., predicate="within")`** which is fast point-in-polygon lookup; matches most of points in seconds.
2. **`sjoin_nearest`** then only on small number of unmatched points.

#### 4. Skip Threshold for Region Re-computation
Also, added a threshold (`> 10` missing regions) to skip re-enrichment entirely when only a handful of samples lack region labels (usually ≤5 after the first run).

---

### Exps on ~7M-row Dataset till Training Start

| Scenario | Time to Training |
|---|---|
| Fresh wide parquet (no regions, row group optimizations) | ~40 min |
| Fresh wide parquet (with optimized region assignment) | **~8 min** |
| Pre-computed regions (subsequent runs) | **~5 min** |

### DIFF
--- a/src/worldcereal/train/data.py
+++ b/src/worldcereal/train/data.py
@@ -108,6 +108,10 @@ def _attach_regions_from_boundaries(
         )
 
     if valid_mask.any():
+        n_target = int(valid_mask.sum())
+        logger.info(
+            f"Assigning regions for {n_target:,} samples via spatial join ..."
+        )
         coords = pd.DataFrame(
             {"lon": lon[valid_mask], "lat": lat[valid_mask]},
             index=df.index[valid_mask],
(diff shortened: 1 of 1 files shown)


src/worldcereal/train/data.py
```diff
@@ -1149,8 +1181,9 @@ def get_training_dfs_from_parquet(
         ).str.strip().eq("")
     else:
         region_missing_mask = pd.Series(True, index=df.index, dtype=bool)
-
-    should_enrich_regions = region_missing_mask.any() and (
+    # only if region_missing_mask has more than 10 missings
```
kvantricht, line 1184: feels like cherry-picking number. What if we only have 50 samples? Then 20% would be missing?

src/worldcereal/train/data.py
```diff
@@ -1162,7 +1195,20 @@ def get_training_dfs_from_parquet(
...
         if not is_tempfile:
-            df.to_parquet(wide_parquet_output_path, index=False)
+            tbl = pa.Table.from_pandas(df, preserve_index=False)
+            pq.write_table(
+                tbl,
+                wide_parquet_output_path,
+                compression="snappy",
+                row_group_size=20_000,
+            )
+            del tbl
```
kvantricht, line 1205: Truly needed here?

src/worldcereal/train/data.py
uceshaf: For now, we only have 5 samples that don't get assigned a region. If we keep it as .any(), it still goes into this just for those 5 samples which it would again not find. So, better to have a low number like 10, above which it should only retry region assignments.

src/worldcereal/train/data.py
uceshaf: Better to save everything with row_groups, even temp files or the (train, test, val) parquet files are also now saved with row groups. Helps even outside of finetuning in reading it easily without going OOM.

kvantricht: I meant to refer just to the `del tbl`

src/worldcereal/train/data.py
kvantricht: could we consider a percentage of the total number of samples instead of a an absolute number? Say if <=1% of samples has a missing region?

src/worldcereal/train/data.py
uceshaf: We are trying to free up memory so I guess as soon as something isn't needed, better to del it to not get into OOM issues.

kvantricht (approved):

Issue #388, opened 2025-07-21
Add mask_cropland parameter to crop type UDP


jdegerickx:
The crop type inference workflow has received a slight update.
From now on, users should be able to dictate whether:
- we first run the default cropland model and use the resulting map as a crop mask for their default model (DEFAULT)
- we just apply their custom model and don't do any masking

A new parameter was introduced (`mask_cropland`, default True) --> https://github.com/WorldCereal/worldcereal-classification/blob/main/src/worldcereal/parameters.py#L128

Can you update the UDP so it can take this as an additional parameter?
Note that this will also affect the number of output products generated by the pipeline. 
Default is 4, if `mask_cropland` == False, there will be only 2.

Example of a run on CDSE in which this mask_cropland parameter was included (and set to True): 
j-25072111581645b0b224be31b860c115

PR #676, opened 2026-04-20
582 create pytoch udps

### LINKED ISSUE #582: Create WorldCereal Pytorch UDPs

VincentVerelst:
- Moved the UDP section from `src` to `scripts`.
- Created two scripts that automatically generate new crop extent and crop type UDPs respectively from the worldcereal source code
- Created two scripts that runs these UDPs for testing purposes
- Two new Pytorch UDPs are ready to be published (after merging to main), once approved. Please check the respective UDPs and the scripts on how to run them, which parameters are needed, and what the default values are for optional parameters.


jdegerickx:
The two outdated run UDP scripts should be removed in the frame of this PR:
scripts/inference/cropland_mapping_udp.py
scripts/inference/croptype_mapping_udp.py

scripts/udp/run_udp_extent.py
```diff
@@ -0,0 +1,66 @@
...
+
+temporal_extent = ["2020-11-01", "2021-10-31"]
+
+# URL to the seasonal model zip
+# model_url = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/models/PhaseII/MultiTask/presto-prometheo-dualtask-SeasonalMultiTaskLoss-month-augment%3DTrue-balance%3DTrue-timeexplicit%3DTrue-masking%3Denabled-run%3D202601240103.zip"
+
+# I would recommend already running the select_s1_orbit_state_vvvh from GFMap during job database creation
+orbit_state = "DESCENDING"  # "ASCENDING" or "DESCENDING"
+
+season_ids = ["tc-s1"]
```
jdegerickx: crop extent should always run on full 12 months, not a single season.
So the season window can be automatically set to equal the temporal extent provided by the user.
In terms of season id, I usually infer this from the end date's year, just to make it a bit easier. You could do in this case:

season_ids = ["2021"]
season_windows = ["2021": ("2020-11-01", "2021-10-31"])]

@kvantricht, what do you think?

scripts/udp/run_udp_extent.py
```diff
@@ -0,0 +1,66 @@
...
+    "west": minx,
+    "south": miny,
+    "east": maxx,
+    "north": maxy,
+    "crs": f"EPSG:{epsg}",
+}
+
+temporal_extent = ["2020-11-01", "2021-10-31"]
+
+# URL to the seasonal model zip
```
jdegerickx: This requires a bit of discussion on the scope of this crop extent UDP.
We have always had the idea to allow users to run their custom crop extent model, e.g. in the WorldCereal processing hub.
With the current model architecture, this would require two parameters:
 
- seasonal_model_zip --> (optional) URL to finetuned presto, which includes a default cropland head. Should default to global landcover presto model as present in the repository.
- landcover_head_zip --> (optional) URL to a specific land cover head trained by the user, on top of the seasonal model.

Possible to expose these parameters for the user?
@kvantricht, please contradict me if you don't agree...

scripts/udp/run_udp_type.py
```diff
@@ -0,0 +1,67 @@
...
+    "south": miny,
+    "east": maxx,
+    "north": maxy,
+    "crs": f"EPSG:{epsg}",
+}
+
+temporal_extent = ["2020-11-01", "2021-10-31"]
+
+# URL to the seasonal model zip
+model_url = "https://artifactory.vgt.vito.be/artifactory/auxdata-public/worldcereal/models/PhaseII/MultiTask/presto-prometheo-dualtask-SeasonalMultiTaskLoss-month-augment%3DTrue-balance%3DTrue-timeexplicit%3DTrue-masking%3Denabled-run%3D202601240103.zip"
```
jdegerickx: For crop type it's even more complex, as we could have 3 models here:

- seasonal_model_zip --> finetuned presto, basis for everything
- landcover_head_zip --> model to be used for cropland masking
- croptype_head_zip --> model to be used for actual crop type predictions

scripts/udp/run_udp_type.py
```diff
@@ -0,0 +1,67 @@
...
+
+# I would recommend already running the select_s1_orbit_state_vvvh from GFMap during job database creation
+orbit_state = "DESCENDING"  # "ASCENDING" or "DESCENDING"
+
+season_ids = ["tc-s1"]
+season_windows = {
+    "tc-s1": ("2020-12-01", "2021-07-31")
+}
+
+postprocess_method = "majority_vote"  # "majority_vote" or "smooth_probabilities", defaults to "majority_vote" if not provided
```
jdegerickx: I guess we agreed to have same postprocessing options for both cropland AND crop type?
Because in our user-facing workflows, we allow them to set these separately.

scripts/udp/run_udp_type.py
```diff
@@ -0,0 +1,67 @@
...
+
+# -------------------------------------------------------------------------------
+
+# -------------------------------------------------------------------------------
+# CREATE OPENEO PROCESS GRAPH FROM UDP AND RUN IT
+# -------------------------------------------------------------------------------
+
+c = openeo.connect("https://openeo.dataspace.copernicus.eu").authenticate_oidc()
+
+inference_cube = c.datacube_from_process(
```
jdegerickx, line 50: so this UDP would not allow users to enable/disable cropland masking?

Normally we expose two parameters to control this behaviour:
- enable_cropland_head --> if true, this means the cropland product will be generated together with the crop type product
- enforce_cropland_gate --> if true, this means the crop type product will be masked by the cropland product

jdegerickx: You can have a look in this notebook to see what I mean: https://github.com/WorldCereal/worldcereal-classification/blob/main/notebooks/trainings/worldcereal_custom_croptype.ipynb

jdegerickx (commented):
Added some comments based on what I would expect should be possible using the UDP's.

Based on our conversation with Thomas, I understood that in the future all WorldCereal map generation requests done by users should actually be routed through these UDP's, so I feel like in that case we should add some more parameters to the UDP to be fully compatible with our current workflows.

@kvantricht, please add your view as well

scripts/udp/run_udp_extent.py
kvantricht: In theory we could also plugin the custom head into the seasonal_model so we only need to pass one model in such case. But for simplicity and flexibility, landcover_head_zip as optional parameter could be interesting indeed.

scripts/udp/run_udp_extent.py
kvantricht: sounds good!

scripts/udp/run_udp_type.py
kvantricht: That's correct.

scripts/udp/run_udp_type.py
kvantricht: Indeed something you have requested long time ago already: https://github.com/WorldCereal/worldcereal-classification/issues/388
Would be good to put this in now, maybe with updated more intuitive name as discussed.

VincentVerelst:
@kvantricht , @jdegerickx , thanks for the feedback. I understand it's mostly a matter of adding some extra optional parameters. To avoid all confusion, I'll list my interpretation of the parameters required for each UDP. Can you confirm this is correct, or correct me where I'm wrong?

**worldcereal_crop_extent**

| Parameter | Optional? | Default |
|----------|----------|----------|
| spatial_extent  | NO  | /  |
| temporal_extent  | NO  | /  |
| orbit_state  | NO  | /  |
| postprocess_method  | YES  | majority_vote  |
| postprocess_kernel_size  | YES  | 5  |
| season_ids  | NO  | /  |
| season_windows  | NO  | /  |
| seasonal_model_zip | YES |https://s3.waw3-1.cloudferro.com/project_dependencies/worldcereal/presto-prometheo-dualtask-SeasonalMultiTaskLoss-month-augment=True-balance=True-timeexplicit=True-masking=enabled-run=202601240103.zip |
| landcover_head_zip | YES | None |

**worldcereal_crop_type**

| Parameter | Optional? | Default |
|----------|----------|----------|
| spatial_extent  | NO  | /  |
| temporal_extent  | NO  | /  |
| orbit_state  | NO  | /  |
| postprocess_method_cropland  | YES  | majority_vote  |
| postprocess_kernel_size_cropland  | YES  | 5  |
| postprocess_method_croptype  | YES  | majority_vote  |
| postprocess_kernel_size_croptype  | YES  | 5  |
| season_ids  | NO  | /  |
| season_windows  | NO  | /  |
| seasonal_model_zip | YES |https://s3.waw3-1.cloudferro.com/project_dependencies/worldcereal/presto-prometheo-dualtask-SeasonalMultiTaskLoss-month-augment=True-balance=True-timeexplicit=True-masking=enabled-run=202601240103.zip |
| landcover_head_zip | YES | None |
| croptype_head_zip | YES | None |
| enable_cropland_head | YES | True |
| mask_cropland | YES | True |

jdegerickx:
> @kvantricht , @jdegerickx , thanks for the feedback. I understand it's mostly a matter of adding some extra optional parameters. To avoid all confusion, I'll list my interpretation of the parameters required for each UDP. Can you confirm this is correct, or correct me where I'm wrong?
> 
> **worldcereal_crop_extent**
> 
> Parameter	Optional?	Default
> spatial_extent	NO	/
> temporal_extent	NO	/
> orbit_state	NO	/
> postprocess_method	YES	smooth_probabilities
> postprocess_kernel_size	YES	5
> season_ids	NO	/
> season_windows	NO	/
> seasonal_model_zip	YES	https://s3.waw3-1.cloudferro.com/project_dependencies/worldcereal/presto-prometheo-dualtask-SeasonalMultiTaskLoss-month-augment=True-balance=True-timeexplicit=True-masking=enabled-run=202601240103.zip
> landcover_head_zip	YES	None
> **worldcereal_crop_type**
> 
> Parameter	Optional?	Default
> spatial_extent	NO	/
> temporal_extent	NO	/
> orbit_state	NO	/
> postprocess_method	YES	smooth_probabilities
> postprocess_kernel_size	YES	5
> season_ids	NO	/
> season_windows	NO	/
> seasonal_model_zip	NO	/
> landcover_head_zip	YES	None
> croptype_head_zip	YES	None
> enable_cropland_head	YES	True
> enforce_cropland_gate	YES	True

The seasonal_model_zip in the crop TYPE UDP should also default to the same one used in the crop extent UDP.
Other than that should be fine

kvantricht:
> > @kvantricht , @jdegerickx , thanks for the feedback. I understand it's mostly a matter of adding some extra optional parameters. To avoid all confusion, I'll list my interpretation of the parameters required for each UDP. Can you confirm this is correct, or correct me where I'm wrong?
> > **worldcereal_crop_extent**
> > Parameter	Optional?	Default
> > spatial_extent	NO	/
> > temporal_extent	NO	/
> > orbit_state	NO	/
> > postprocess_method	YES	smooth_probabilities
> > postprocess_kernel_size	YES	5
> > season_ids	NO	/
> > season_windows	NO	/
> > seasonal_model_zip	YES	https://s3.waw3-1.cloudferro.com/project_dependencies/worldcereal/presto-prometheo-dualtask-SeasonalMultiTaskLoss-month-augment=True-balance=True-timeexplicit=True-masking=enabled-run=202601240103.zip
> > landcover_head_zip	YES	None
> > **worldcereal_crop_type**
> > Parameter	Optional?	Default
> > spatial_extent	NO	/
> > temporal_extent	NO	/
> > orbit_state	NO	/
> > postprocess_method	YES	smooth_probabilities
> > postprocess_kernel_size	YES	5
> > season_ids	NO	/
> > season_windows	NO	/
> > seasonal_model_zip	NO	/
> > landcover_head_zip	YES	None
> > croptype_head_zip	YES	None
> > enable_cropland_head	YES	True
> > enforce_cropland_gate	YES	True
> 
> The seasonal_model_zip in the crop TYPE UDP should also default to the same one used in the crop extent UDP. Other than that should be fine

**crop extent**
- Default postprocess_method should probably be majority voting; kernel size should normally be 3 (for smallholder farming etc). Is the problem here that we cannot have separate settings for crop extent and crop type?
- Season IDs and windows: actually makes the UDP quite noisy because strictly speaking we don't have a "season", just a year; should we adapt this so the UDP does not take seasons in this case?

**crop type**
- Default postprocess_method should here also be majority voting; kernel size 5 is what turned out to be best in phase I. but not sure if we can divert from what we do for cropland? Unless we just put the parameter with respect to some MMU. 
- enable_cropland_head -> how can this be False in croptype UDP?
- enforce_cropland_gate -> rename to mask_cropland?

jdegerickx:
@VincentVerelst (CC: @kvantricht)

I did the refactor renaming enforce_cropland_gate to mask_cropland everywhere.
I ran the script to generate the UDP's and also pushed the resulting json files to this branch.
I also adjusted the links in the run_udp scripts (they were still pointing to src) and tested both UDPs.
Both generated valid outcomes.

Regarding postprocessing ideally we indeed have the possibility to have different settings for cropland vs crop type.

I thought enable_cropland_head can be False in crop type UDP if you don't wish to mask cropland nor generate the cropland mask? So I guess fine?

kvantricht:
> enable_cropland_head

ah yes got confused there. But if mask_cropland is True and this is set to False, there's a problem. What's the typical use case here? We could also opt to only expose the mask flag and if True, cropland product is saved always.

jdegerickx:
> > enable_cropland_head
> 
> ah yes got confused there. But if mask_cropland is True and this is set to False, there's a problem. What's the typical use case here? We could also opt to only expose the mask flag and if True, cropland product is saved always.

There is some internal logic present which overrides enable_cropland_head automatically in case mask_cropland is set to True

jdegerickx:
@VincentVerelst, I updated the parameter table in this issue --> https://github.com/WorldCereal/worldcereal-vdm/issues/32

VincentVerelst:
@jdegerickx , the diff has become quite large in this PR, but a lot is just your renaming of the `cropland_mask` parameter. 

The most important parts are under `/scripts/udp/`, which contains scripts to make the UDPs, scripts to run the UDPs, and the UDPs themselves.

scripts/udp/make_udp_type.py
```diff
@@ -0,0 +1,489 @@
...
+            "enabled": True,
+            "method": CROPTYPE_DEFAULTS["method"],
+            "kernel_size": CROPTYPE_DEFAULTS["kernel_size"],
+        }
+
+    workflow_cfg = WorldCerealWorkflowConfig(
+        model=ModelSection(
+            export_embeddings=export_embeddings,
+            export_ndvi=export_ndvi,
+            enable_cropland_head=True,
```
jdegerickx, line 103: enable_cropland_head is hardcoded here to be always True?
Should be adjustable based on the parameter, not sure this is the case...

scripts/udp/make_udp_type.py
```diff
@@ -0,0 +1,489 @@
...
+    if isinstance(obj, dict):
+        for key, value in obj.items():
+            if key == "landcover_head_zip" and value == target_landcover_head_zip:
+                obj[key] = {"from_parameter": "landcover_head_zip"}
+            else:
+                replace_landcover_head_zip(value, target_landcover_head_zip)
+    elif isinstance(obj, list):
+        for item in obj:
+            replace_landcover_head_zip(item, target_landcover_head_zip)
+
```
jdegerickx, line 279: missing a parameter for a custom crop type head zip (croptype_head_zip) --> see table https://github.com/WorldCereal/worldcereal-vdm/issues/32

jdegerickx (changes requested):
almost there, I believe!

jdegerickx:
@VincentVerelst, I quickly added all optional parameters in the run_udp scripts so that it would be clear for GISAT how to parameterize them fully. Hope you don't mind :)

scripts/udp/make_udp_type.py
VincentVerelst: Yes, the script works in two parts: 

1. Create a process graph without any parameters, and set some random values
2. Parametrize everything and set defaults

scripts/udp/make_udp_type.py
VincentVerelst: Ah missed that, added now!

jdegerickx (approved):
We performed quite a bit of tests covering all parameter combinations.
Everything seems to work as expected.

PR #677, opened 2026-04-20
Inference fixes

cbutsko:
Two things related to inference are addressed in this PR:
1. **Device consistency in `_run_batches`.** When GPU is available, predictor payloads are pushed to cuda via `to_torchtensor(..., device=self.device)`. That collided with prometheo's `dataset_to_model`, which assembles the model input on host memory (numpy buffers + slice assignments), causing `TypeError: can't convert cuda:X device type tensor to numpy`. **Fix:** keep predictor fields on CPU and defensively demote any CUDA tensor that sneaks in; only season_masks (consumed directly by the seasonal head) is placed on self.device. The consolidated GPU transfer now happens in exactly one place  - inside the Presto wrapper's forward - which already owns device placement.

**UPD:** next part not relevant anymore, since that was fixed in another place, so the change is reverted:

> This was a bug which is addressed in https://github.com/WorldCereal/worldcereal-finetune/pull/12 of the finetune repo. After that fix, head types will be correctly written to their respective sections and the inference code will work as intended. 

2. **Head architecture resolution from the manifest**. Seasonal dualtask manifests head_type / head_hidden_dim / head_dropout into the backbone section (since both heads share the same architecture), while single-head downstream manifests repeat them per head. `_select_head_spec` only understood the per-head layout and silently fell back to `"linear"/256/0.0`, which caused state_dict mismatches when loading checkpoints trained with MLP heads. **Fix:** `_select_head_spec` now accepts an optional backbone dict and falls back to backbone-level values when a head entry lacks them. Per-head fields still win when present, so both manifest conventions load identically and no repackaging is needed.


kvantricht:
@cbutsko regarding part 2 of your fix: actually this is related to what I was mentioning this morning causing a bug in inference. Upon investigation, it became clear that this head type should not at all be part of the backbone section in the config. This was a bug which is addressed in [this PR](https://github.com/WorldCereal/worldcereal-finetune/pull/12) of the finetune repo. After that fix, head types will be correctly written to their respective sections and the inference code will work as intended.

cbutsko:
> @cbutsko regarding part 2 of your fix: actually this is related to what I was mentioning this morning causing a bug in inference. Upon investigation, it became clear that this head type should not at all be part of the backbone section in the config. This was a bug which is addressed in [this PR](https://github.com/WorldCereal/worldcereal-finetune/pull/12) of the finetune repo. After that fix, head types will be correctly written to their respective sections and the inference code will work as intended.

Great! then I'll revert this change to be as in the main branch.

kvantricht (approved):
Looking good, thanks!

Issue #670, opened 2026-04-15
Possible improvements classification app


kvantricht:
After running through classification app, I though about following possible optimizations:
- launch a referendum about intuitiveness of date slider ticks
- mandatory season ID setting in inference: can we not default to the mandatory season ID setting during training? Or in fact even completely omit this here because it would be weird if a user takes a different name here.
- Can Season slider for inference default to the setting set in an earlier step by the user? If needed adjusted to latest year. But now the season he picked is textually suggested but not applied to the slider by default;
- Product Type in inference defaults to cropland, which feels weird as the entire app is about crop type training. Would suggest to default to croptype product in case we follow the full flow of training crop type model.

jdegerickx:
additional one: allow users to easily re-start map production based on a certain output directory

jdegerickx:
all issues have been addressed in the PR mentioned above, EXCEPT for the intuitiveness of the date slider, as the referendum launched within the WorldCereal chat was inconclusive.

kvantricht:
Closing for now as only the inconclusive referendum stands.

PR #682, opened 2026-04-28
App optimizations

jdegerickx:
Addressing the items raised in #670 

- default product type for custom workflow is crop type
- date slider set automatically to the season for which the model has been trained (REQUIRES UPDATE OF SRC!)
- user has possibility to resume map generation from an existing output directory

Note: Haven't changed anything to the way the ticks are shown in the date slider, as the poll was inconclusive.


kvantricht (approved):
Can't test the app itself right now, but description looks good. No comments on the `src` change but will indeed require a new build.

PR #683, opened 2026-04-29
S2 masking fix

cbutsko:
**BUG:** `disable_s2` and the `s2_cloud_timestep_prob` augmentation were silently no-ops: every "S2-disabled" or "S2-cloudy-timestep" run actually trained on full Sentinel-2.

Per-timestep S2 masking guarded its operation by checking band B1 of S2. but B1 is never populated by the data loader, so its check `band != NODATAVALUE` always returned False.

**Scope:** every `--disable_s2` ablation result is invalid. Default training used weaker S2 cloud augmentation than the config advertised (only the contiguous-block path worked).


kvantricht (approved):
Excellent catch!

PR #686, opened 2026-04-29
Use prometheo==v0.0.5 with ndvi mask fix

kvantricht:
Due to a recently discovered bug in prometheo, NDVI was not properly masked in Presto. This bug has been addressed and is part of release v0.0.5; it's crucial all new workflows use this version of the library instead.

I have updated the reference to the new wheel on s3; **importantly**, UDPs will have to be updated as well! cc @VincentVerelst


VincentVerelst (approved):
Looks good. In my new PR on making new UDPs, the UDPs anyway use the `DEFAULT_INFERENCE_JOB_OPTIONS`, so this should propagate through, once I pull these changes into that feature branch

PR #687, opened 2026-04-29
disable spatial split by default for user-facing training

jdegerickx:
Based on experience from Shabri, it might be best to by default **disable** spatial splitting of cal/val/test for user-facing training workflows.
Outside of Europe we usually have very clustered availability of crops, often leading to certain classes being completely excluded from the TEST set if spatial splitting is enabled.

### DIFF
--- a/notebooks/notebook_utils/classification_app.py
+++ b/notebooks/notebook_utils/classification_app.py
@@ -152,9 +152,7 @@ class WorldCerealClassificationApp:
         self.tab8_widgets: Dict[str, Any] = {}
         self.tab9_widgets: Dict[str, Any] = {}
 
-        display(
-            HTML(
-                """
+        display(HTML("""
                 <style>
                     .jp-OutputArea-output pre {
                         white-space: pre-wrap !important;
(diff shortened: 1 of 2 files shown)


kvantricht:
I've seen things like that as well; much depends though on size of the spatial bins, which could already help. So another option could be reduce the default size instead, but maybe not worth the additional try?

jdegerickx:
sounds a bit risky as how will we determine a good default value which is guaranteed to reasonably work well in many cases?
Perhaps the default path should be as easy as possible with a relatively high chance of getting good metrics? :)

kvantricht:
As long as you're OK with the optimistic bias reflected by those metrics :D Just keep in mind that by switching it off, you could have in such cases that multiple people recording the same field results in samples getting spread around train/val/test and this will blow up the metrics for sure. Agreed that sometimes it can hurt too. By the way, risky is a big word; I assume that a spatial bin size of 0.00001 eventually results in the same kind of splitting as you would achieve by switching it off. It just allow to specify e.g. a bin size of a typical field so somehow some sense of auto-correlation of neighboring samples is avoided.

jdegerickx:
understood.
I don't have a clear preference here tbh, i.e. switching it off or lower the bin size.
In the end both options might have more or less the same result, as we would need to lower the bin size that drastically that it in the end will resemble some kind of stratified random sampling :)

kvantricht (approved):
OK then let's disable it by default indeed for now.

PR #690, opened 2026-04-30
Spatially-aware class balancing for DualHeadBatchSampler

cbutsko:
Adds an optional per-bin class-weighting mode to the dual-head batch sampler so class weights can be derived locally per lat/lon bin instead of once globally. Default behaviour is unchanged.

**What's new**
* `_get_per_bin_class_weights`: computes class weights within each lat/lon bin via the existing `get_class_weights` machinery. Bins below a configurable `min_samples_per_bin` (default 50) fall back to the globally-computed weights. 
* `_get_spatial_density_weights`: wraps the existing density-factor computation with a sparse-bin override: bins below `min_samples_per_bin` get weight = 1.0 instead of an extreme inverse-density value. Without this, singleton bins (e.g. one sample in a remote ocean cell) pin the density factor to its upper clip and blow up the multiplicative composition with class weights.
* `DualHeadBatchSampler` refactor: exposes two new orthogonal axes:
    * `class_balancing_scope: Literal["global", "per_bin"]` - source of class weights.
    * `spatial_weight_method` (existing) controls the density factor independently and can be set to "none" to skip it.
    * Both axes compose multiplicatively. Useful combinations include per_bin × none (within-bin class balance only) and per_bin × balanced (within-bin balance + cross-bin density equalisation).

**Backward compatibility**
Default `class_balancing_scope="global"` reproduces the previous behaviour exactly.

**UPD:**
* Added  inspectable artifacts so we can sanity-check what the sampler actually draws with - useful for debugging sparsity, fallbacks, and density effects when using spatial bin based balancing.
* New module: `worldcereal/train/diagnostics.py`. Standalone home for training-time inspection helpers; first occupant is `dump_per_bin_class_grids`. Keeps `datasets.py` focused on data/sampler logic and gives future diagnostics an obvious place to land.

### DIFF
--- a/src/worldcereal/train/datasets.py
+++ b/src/worldcereal/train/datasets.py
@@ -794,6 +794,7 @@ def _get_smoothed_per_bin_class_weights(
             str_labels, lat_bin, lon_bin, method, min_samples_per_bin,
             lat_min, lat_max, lon_min, lon_max,
             min_samples_per_class_per_bin=min_samples_per_class_per_bin,
+            pool_name=pool_name,
         )
     )
     n_lat = grid.shape[1]
@@ -870,6 +871,7 @@ def _build_per_class_weight_grid(
     grid_lat_min: int, grid_lat_max: int,
     grid_lon_min: int, grid_lon_max: int,
     min_samples_per_class_per_bin: Optional[int] = None,
+    pool_name: str = "",
 ) -> Tuple[np.ndarray, Dict[str, int], Dict[str, float], int, int]:
     """Build a (n_classes, n_lat, n_lon) per-class lookup grid.
 
--- a/src/worldcereal/train/datasets.py
+++ b/src/worldcereal/train/datasets.py
@@ -667,6 +667,185 @@ def _get_per_bin_class_weights(
     return weights
 
 
+def _get_smoothed_per_bin_class_weights(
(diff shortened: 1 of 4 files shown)


src/worldcereal/train/datasets.py
```diff
@@ -513,6 +513,44 @@ def _get_normalized_weights(
...
+    the Arctic can pin the density factor to its upper clip),
+    blowing up the multiplicative composition with class weights downstream.
+
+    Mirrors the analogous safeguard in `_get_per_bin_class_weights`:
+    a bin is considered too sparse to estimate a stable density weight from,
+    and is treated as "average density" instead.
+    """
+    sp_arr = _get_normalized_weights(spatial_bins, method, clip_range)
+
+    if min_samples_per_bin > 1:
```
kvantricht, line 557: Should we have a safeguard for invalid values < 0? And what happens with ==1? It switches off the min_samples special treatment? Let's explain this explicitly here.

cbutsko: good one, addressed here https://github.com/WorldCereal/worldcereal-classification/pull/690/commits/ea4ad092c1af84981b347bd78758f8daa2482954

kvantricht (approved):
nice additions! looks ready to merge and test in the real world.

PR #697, opened 2026-05-06
Adding finetune utilities

kvantricht:
### Summary

Two new knobs for dual-head finetuning, plus a ruff formatting pass over `datasets.py`.

---

#### `datasets.py` — per-class weight multipliers

A new optional `class_weight_multipliers: Optional[Mapping[str, float]]` parameter is added to both `DualHeadBatchSampler.__init__()` and `WorldCerealLabelledDataset.get_dual_head_batch_sampler()`.

The dict accepts class labels from **either** the LC or CT label space. Routing is automatic: at construction time the sampler compares each key against the observed LC and CT label sets and applies multipliers to the correct pool. Unknown keys produce a warning and are ignored. This avoids the need for two separate arguments and keeps the interface label-space-agnostic.

The multipliers are applied **after** the main class-balancing weights (global or per-bin), so they compose multiplicatively on top of whatever balancing scheme is already in use.

#### finetuning_utils.py — flexible checkpoint metric

A new `checkpoint_metric: Literal["val_loss", "lc_f1", "ct_f1", "mean_f1"]` parameter (default `"val_loss"`) is added to `run_finetuning()`. It controls which metric drives early stopping and best-checkpoint selection:

| value | behaviour |
|---|---|
| `val_loss` | save when total validation loss decreases (previous behaviour) |
| `lc_f1` | save when validation landcover macro F1 increases |
| `ct_f1` | save when validation crop-type macro F1 increases |
| `mean_f1` | **(new default)** save when the average of LC and CT macro F1 increases |

`best_loss` is always kept in sync regardless of the chosen metric, so existing callbacks and `_save_best` continue to receive a valid loss value.

#### Formatting

`datasets.py` received a ruff formatting pass (argument lists reflowed to one-per-line, no logic changes).

### DIFF
--- a/src/worldcereal/train/datasets.py
+++ b/src/worldcereal/train/datasets.py
@@ -2089,6 +2089,7 @@ class WorldCerealLabelledDataset(WorldCerealDataset):
         min_samples_per_class_per_bin: Optional[int] = None,
         num_batches: Optional[int] = None,
         generator: Optional[torch.Generator] = None,
+        class_weight_multipliers: Optional[Mapping[str, float]] = None,
     ) -> "DualHeadBatchSampler":
         """Build a :class:`DualHeadBatchSampler` for dual-head training.
 
@@ -2111,6 +2112,7 @@ class WorldCerealLabelledDataset(WorldCerealDataset):
             min_samples_per_class_per_bin=min_samples_per_class_per_bin,
             num_batches=num_batches,
             generator=generator,
+            class_weight_multipliers=class_weight_multipliers,
         )
 
 
@@ -2183,6 +2185,7 @@ class DualHeadBatchSampler(Sampler):
         min_samples_per_class_per_bin: Optional[int] = None,
         num_batches: Optional[int] = None,
         generator: Optional[torch.Generator] = None,
+        class_weight_multipliers: Optional[Mapping[str, float]] = None,
     ) -> None:
         import math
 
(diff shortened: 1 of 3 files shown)


uceshaf:
There are a couple of bugs and one small improvement, I will push a commit before merge, if you agree @kvantricht ?

- **cur_mean_f1 potentially corrupted by -1.0 sentinel** (finetuning_utils.py)

The arithmetic mean may becomes nonsensical and do incorrect checkpoint selection / premature early stopping under the new default checkpoint_metric="mean_f1" if CT rejection empties seasonal_metrics_flat for a given epoch (less likely but maybe possible in regional runs).

**Before**:
```
cur_lc_f1 = seasonal_metrics_flat.get("landcover/f1_macro", -1.0)
cur_ct_f1 = seasonal_metrics_flat.get("croptype/f1_macro", -1.0)  
cur_mean_f1 = (cur_lc_f1 + cur_ct_f1) / 2.0
```

**After**:
```
cur_lc_f1 = seasonal_metrics_flat.get("landcover/f1_macro", -1.0)
cur_ct_f1 = seasonal_metrics_flat.get("croptype/f1_macro", -1.0)
cur_mean_f1 = (max(0.0, cur_lc_f1) + max(0.0, cur_ct_f1)) / 2.0
```

- **elif silently drops CT multiplier for labels that are in both LC and CT sets** (datasets.py)

Labels like "temporary_crops" can appear in both lc_label_set and ct_label_set. The elif means the CT will never receives its multiplier.

**Before**:
```
for key, val in class_weight_multipliers.items():
    if key in lc_label_set:
        lc_multipliers[key] = val
    elif key in ct_label_set:
        ct_multipliers[key] = val
    else:
        logger.warning(...)
```
**After**:
```
for key, val in class_weight_multipliers.items():
    found = False
    if key in lc_label_set:
        lc_multipliers[key] = val
        found = True
    if key in ct_label_set:
        ct_multipliers[key] = val
        found = True
    if not found:
        logger.warning(...)
```

-  **seasonal_metrics_supported / _effective_metric recomputed each epoch** (finetuning_utils.py)

Both depend only on loss_fn (constant for the run). Moving them above the epoch loop avoids redundant work and is perhaps clearer.

**Before**: both computed inside the for epoch in range(...) / validation block.
**After**: put seasonal_metrics_supported (and derive _effective_metric) once before the loop.

kvantricht:
Sounds good; in practice i don't think "temporary_crops" is something that will also occur in crop types. In my iteration it was intentionally decided that there cannot be overlap. But who knows, maybe there can be in some distant future. Better safe than sorry, so agreed. Should we also add a test on that metric monitoring behavior? E.g. the effective_metric addition was done to fix a test, because in non-seasonal models this was already giving issues.

uceshaf:
**Re: fixes:** Pushed the changes in one commit
 
**Re: tests:** added two new test classes to test_datasets.py:

- **TestClassWeightMultipliersRouting** (3 tests): verifies that a label in both pools boosts both, a LC-only label doesn't touch CT, and an unknown label doesn't crash.
- **TestCheckpointMetricMonitoring** (7 tests): verifies all four mean_f1 sentinel cases (both available, LC sentinel, CT sentinel, both sentinel → 0.0) and the seasonal/non-seasonal _effective_metric fallback paths.

All 10 new tests pass. Full dataset suite: 92 passed.

Just waiting for 1 complete epoch to finish now.

uceshaf (approved):

Issue #710, opened 2026-06-01
Fix filename prefix in UDPs


VincentVerelst:
The filename prefixes are hardcoded in the UDP. Best to give them a more proper name like we used to.

PR #712, opened 2026-06-02
Set meaningful filename_prefix in inference UDPs #710

VincentVerelst:


jdegerickx (commented):
To me this PR triggers the question: is there a way to make the job title more dynamic, i.e. including information about spatial and temporal extent?

For instance, in the notebooks we use the following:

title=f"WorldCereal [{product_type.value}] job_{row.tile_name}"
--> WorldCereal [croptype] job_34TBK_0_0

jdegerickx (approved):

PR #715, opened 2026-06-11
Finetune updates

uceshaf:
Mostly region related updates:
* Error when a region in list of requested regions doesn't match
* Update definition of some regions (individual commits show changes)
* Option to force recompute regions even if already in the merged parquet files


cbutsko (approved):
Looks good!

PR #721, opened 2026-06-19
Consistent season-window date handling (fixes silent in-season sample drops)

cbutsko:
**Problem**
Crop classes that were well represented in the extractions could end up with near-zero samples after training (e.g. wheat: ~6700 → ~18), surfacing only in the confusion matrix. Root cause: a sample's `in_season` flag is computed against the season-window end trimmed down to the start of its composite (`Aug 31 → Aug 01`), while the exact `valid_time` is compared against it. Any label in the second half of the window's final month (e.g. Aug 15) was marked out-of-season and silently dropped late in `_drop_invalid_samples` - after embeddings were computed.

This was made worse by three places asking _"is this date in the season window?"_ with inconsistent logic: classification app the Tab-3 alignment pre-filter and the manual-window dataset filter used the actual end date (lenient), while the in_season flag used the floored end (strict). So a sample passed alignment, got embeddings, then disappeared.

**Fix**
A single, composite-aware membership rule lives in `worldcereal/train/seasonal.py` and is shared by all three call sites: the season start snaps down to its _composite_ start, the season end snaps up to its _composite_ end (`composite_window_end`), and the exact `valid_time` is compared against that range. This matches the season mask the embedding already uses, and works for both monthly and dekadal compositing.

**Changes**
* `seasonal.py` is now the single home for season/composite primitives.
* `refdata.py`: season-window pre-filter uses `in_season_window(..., freq=freq)`; removed the duplicate `_season_window_membership` and `import calendar`.
* `datasets.py`: in_season flag in `_season_mask_from_window`/`_season_mask_from_calendar` now uses `date_in_season` (coverage gating preserved); removed the duplicate definitions, and the `_advance_composite_slot/_enumerate_composite_slots` wrapper methods (they now live in `seasonal.py`). 
* `test_seasonal.py`: new suite covering the grid primitives, window resolution (incl. year-crossing + day-clamping), and membership.

**Net:** 3 duplicated membership implementations → 1; ~170 lines removed.

**Behavior change**
The alignment pre-filter is now composite-aware (it was day-resolution). For whole-month / whole-dekad windows this is identical to before. It only differs for windows whose edges fall mid-composite, where it now rounds to the composite edge, which is the point: the pre-filter and the in_season flag can no longer disagree.

### DIFF
--- a/src/worldcereal/train/datasets.py
+++ b/src/worldcereal/train/datasets.py
@@ -47,7 +47,7 @@ from worldcereal.train import (
 from worldcereal.train import predictors as _predictor_utils
 from worldcereal.train.seasonal import (
     SeasonWindow,
-    _coerce_date_for_year,
+    coerce_date_for_year,
     align_to_composite_window,
     date_in_season,
     enumerate_composite_slots,
@@ -1561,9 +1561,9 @@ class WorldCerealDataset(Dataset):
 
         cycles: List[Tuple[np.datetime64, np.datetime64]] = []
         for year in sorted(base_years):
-            start_dt = _coerce_date_for_year(year, window.start_month, window.start_day)
+            start_dt = coerce_date_for_year(year, window.start_month, window.start_day)
             end_year = year + window.year_offset
-            end_dt = _coerce_date_for_year(end_year, window.end_month, window.end_day)
+            end_dt = coerce_date_for_year(end_year, window.end_month, window.end_day)
             start_aligned = align_to_composite_window(start_dt, self.timestep_freq)
             end_aligned = align_to_composite_window(end_dt, self.timestep_freq)
             if end_aligned < start_aligned:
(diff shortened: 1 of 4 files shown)


src/worldcereal/train/datasets.py
```diff
@@ -46,7 +45,15 @@
     SEASONALITY_LOOKUP_PATH,
 )
 from worldcereal.train import predictors as _predictor_utils
-from worldcereal.train.seasonal import align_to_composite_window
+from worldcereal.train.seasonal import (
+    SeasonWindow,
+    _coerce_date_for_year,
```
kvantricht: this should probably not be a private function if we import elsewhere. Also, should we find a more informative name for this method?

kvantricht: but on second thought, is this still used here?

kvantricht (commented):
Nice! We'll have to find out in a finetuning run if we're as efficient as before when every sample in a batch will have to be processed by the seasonality functions.

kvantricht (approved):
Looks good!

PR #724, opened 2026-07-07
By default mask S2 B8A in inference

kvantricht:
#### Description
This PR adds support for controlling Sentinel-2 B8A band masking during inference to maintain consistency with training data. Several key datasets used for training lacked B8A data, creating a train-test mismatch during inference. To address this, models are now trained with B8A always masked. This change ensures inference also masks B8A by default.

#### Problem
- Training datasets (particularly early phases) did not reliably contain Sentinel-2 B8A (NIR narrow) band data
- During inference, B8A values were passed to models that were trained without seeing this band
- This created a data distribution shift and inconsistency between training and inference
- Models trained with B8A masked perform better on production data

#### Solution
Add a `mask_b8a` flag (default: `True`) that allows control of B8A masking at inference time:

- Default behavior (mask_b8a=True): B8A band values are set to NODATA_VALUE during preprocessing, maintaining consistency with training data
- Optional behavior (mask_b8a=False): B8A band is preserved for use cases requiring it

The flag is implemented as a single control point that works uniformly across both local and online (openEO UDF) inference paths.

#### Changes
- SeasonalInferenceEngine.init(): Added `mask_b8a: bool = True` parameter
- run_seasonal_workflow(): Added `mask_b8a: bool = True` parameter to pass through to engine
- _prepare_array(): Masking applied after band name harmonization, before type conversion (single, safe point in pipeline)
- _extract_udf_configuration() and _finalize_workflow_config(): Extract and validate `mask_b8a from UDF context with default `True`
- Tests: Added `test_prepare_array_masks_b8a_band_when_enabled()` and `test_prepare_array_preserves_b8a_band_when_disabled()` to verify both states

#### Usage
Local inference:

```
datacube = run_seasonal_workflow(
    arr, epsg,
    seasonal_model_zip="path/to/model.zip",
    mask_b8a=False  # Optional: disable masking if needed
)
```

Online inference (openEO UDF context):

```
{
    "runtime": {
        "mask_b8a": False  # or true, defaults to true
    }
}
```

#### Backwards Compatibility
- Fully backwards compatible; default True maintains the intended behavior
- Existing code without the parameter continues to work and masks B8A as intended
- Both local and online inference support the flag uniformly

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -33,4 +33,4 @@ jobs:
         OPENEO_AUTH_CLIENT_ID_CDSE: openeo-worldcereal-service-account
         OPENEO_AUTH_CLIENT_ID_VITO: openeo-worldcereal-service-account
         OPENEO_AUTH_CLIENT_SECRET_CDSE: ${{ secrets.OPENEO_AUTH_CLIENT_SECRET_CDSE }}
-        OPENEO_AUTH_CLIENT_SECRET_VITO: ${{ secrets.OPENEO_AUTH_CLIENT_SECRET_TERRASCOPE }}
+        OPENEO_AUTH_CLIENT_SECRET_VITO: ${{ secrets.OPENEO_AUTH_CLIENT_SECRET_CDSE }}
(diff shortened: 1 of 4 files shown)


kvantricht:
@VincentVerelst merging of this PR will need to be followed by retriggering the UDP creation of both crop extent and crop type (and Gisat needs to be notified). This would add this new parameter with proper default `True` value to the PG (should not be exposed in the UDP at this point).

but let’s only do that once we have a new global default model as well trained without this band so we can add it to the UDP

VincentVerelst (approved):

PR #725, opened 2026-07-08
improved logging for downstream training

kvantricht:
The user workflow now shows improved logging on dropped samples due to out of season, per split.


jdegerickx (commented):
Changes look good, but tests require fixing?

kvantricht:
> Changes look good, but tests require fixing?

Yes, known issue with Terrascope auth now also going through CDSE. Vincent is checking this.

jdegerickx (approved):

PR #728, opened 2026-07-09
bugfix targetyear in wrong year, now based on label_datetime

uceshaf:
refyear was being calculated as the year when the season ended, and it was being done without considering the sample datetime, resulting in dropping of points due to valid_time being out of seasons.


kvantricht (approved):
Looks good if linting error is fixed!

PR #730, opened 2026-07-09
Update masking to always return some s1 or s2 data if possible

gabrieltseng:


src/worldcereal/train/datasets.py
```diff
@@ -1306,13 +1341,57 @@ def _apply_masking(
...
+        data, would leave neither sensor with a single valid timestep, one
+        synthetically dropped timestep is restored. S2 is restored in
+        preference to S1 because the dominant violating path is the explicit
+        full-S1-dropout draw, whose semantics should stay intact. A sensor
+        whose full elimination is intentional (dropout probability of 1.0,
+        e.g. disable_s1/disable_s2 experiments) is never restored.
+        """
+        cfg: SensorMaskingConfig = self.masking_config  # type: ignore[assignment]
+        if not (s1_drop | s1_missing).all() or not (s2_drop | s2_missing).all():
+            return s1_drop, s2_drop
```
gabrieltseng, line 1495: this is where this function should exit most of the time. it only continues if s1 & s2 are completely masked or missing, which is not desirable

src/worldcereal/train/datasets.py
```diff
@@ -1306,13 +1341,57 @@ def _apply_masking(
...
+            (
+                s1_drop,
+                np.flatnonzero(s1_drop & ~s1_missing),
+                cfg.s1_full_dropout_prob >= 1.0,
+            ),
+        ]
+        for drop, restorable, intentional in candidates:
+            if restorable.size and not intentional:
+                drop[np.random.choice(restorable)] = False
+                return s1_drop, s2_drop
```
gabrieltseng, line 1512: currently, only a single valid timestep (for s1 or s2, not both) is returned (that's why there's the `return` right after the mask is updated). I don't know if this is desirable but I also expect this condition to be triggered relatively rarely so its probably ok.

src/worldcereal/train/datasets.py
```diff
@@ -1050,11 +1061,33 @@ def __init__(
...
+        how masking is drawn.
+        """
+        s1_cols = [c for c in self.dataframe.columns if c.startswith("SAR-")]
+        s2_cols = [c for c in self.dataframe.columns if c.startswith("OPTICAL-")]
+        if not s1_cols or not s2_cols:
+            return
+        s1_gone = (self.dataframe[s1_cols] == NODATAVALUE).all(axis=1)
+        s2_gone = (self.dataframe[s2_cols] == NODATAVALUE).all(axis=1)
+        num_bad = int((s1_gone & s2_gone).sum())
+        if num_bad:
```
kvantricht: should we also report the percentage in the dataset to have a better feeling of the relative size of the issue?

gabrieltseng: added in the logging below:

```
f"{num_bad}/{len(self)} sample(s) have no S1 and no S2 data at all; "
```

PR #736, opened 2026-07-16
Regional mean f1 + clamping/multipliers fix

cbutsko:
Adding computation of regional mean f1 and possibility to use it as a stopping metric.

### DIFF
--- a/src/worldcereal/train/finetuning_utils.py
+++ b/src/worldcereal/train/finetuning_utils.py
@@ -560,7 +560,6 @@ def _log_regional_metrics(
     records: List[dict],
     task_name: str,
     class_names: Optional[List[str]] = None,
-    *,
 ) -> None:
     """Log per-region macro F1 summary derived from prediction records.
 
--- a/src/worldcereal/train/finetuning_utils.py
+++ b/src/worldcereal/train/finetuning_utils.py
@@ -535,10 +535,32 @@ def _records_to_regional_mean_f1(
     return float(np.mean(regional_f1s))
 
 
+_CROP_FPR_POSITIVE_CLASS = "temporary_crops"
+
+
+def _crop_false_positive_ratio(
+    y_true: Sequence[str],
+    y_pred: Sequence[str],
+    crop_class: str = _CROP_FPR_POSITIVE_CLASS,
+) -> Optional[float]:
+    """Share of true non-crop samples predicted as ``crop_class`` (FPR).
+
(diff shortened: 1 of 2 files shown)


src/worldcereal/train/finetuning_utils.py
```diff
@@ -494,6 +494,68 @@ def _records_to_scalar_metrics(records: List[dict]) -> dict[str, float]:
...
+            )
+        except Exception as exc:  # noqa: BLE001
+            logger.warning(f"Failed computing macro F1 for region '{region}': {exc}")
+
+    if len(regional_f1s) < 2:
+        return None
+    return float(np.mean(regional_f1s))
+
+
+_CROP_FPR_POSITIVE_CLASS = "temporary_crops"
```
kvantricht: this is hardcoded while it's normally set at the finetune level?

src/worldcereal/train/finetuning_utils.py
```diff
@@ -1036,7 +1100,7 @@ def forward(
         loss = torch.zeros(1, device=device, dtype=torch.float32)
 
         # Landcover pathway – only activate for samples explicitly routed to the LC
-        # head (tasks[idx] == landcover_task_name).  When the SeasonalTaskBatchSampler is
+        # head (tasks[idx] == landcover_task_name).  When the DualHeadBatchSampler is
```
kvantricht: don't get why this is renamed back to DualHeadBatchSampler ?

cbutsko: not sure how that slipped in... fixed here https://github.com/WorldCereal/worldcereal-classification/pull/736/changes/e9e5369c74336e03b5101420e845fc972d374cfb

src/worldcereal/train/finetuning_utils.py
cbutsko: yeah, good point. It was just a practical decision to hardcode it here. it's monitoring-only, and there's a very specific type of commission I wanted to add - non_cropland predicted as temporary_crops. 
Looking deeper, though, we already have that concept in this very file. So I've rewired the FPR function to reuse the cropland definition set at finetuning time (`cropland_class_names`) instead of assuming one here https://github.com/WorldCereal/worldcereal-classification/pull/736/changes/e26c82b74a29cb4030f967d1603b03735aa9c76c

kvantricht (approved):
Ready for testing :-)

PR #737, opened 2026-07-16
Europe calendars

jdegerickx:
seasonality_lookup.parquet was updated:
- new crop calendars for Europe (script for corrections will be uploaded to globalproduction repo)
- to have more columns in there with not only doy notation but also real date and dekad numbers, to make it easier to update e.g. the production grid.
- also a unique point_id was added so future updates can be done more easily

Functionality to read the seasonality_lookup.parquet was moved away from the "train" part of the repository and centralized in the cropcalendar section, so it can be more easily re-used across other parts of the repo


kvantricht (approved):
Thanks Jeroen, all good now!

PR #740, opened 2026-07-21
Local inference consolidation

cbutsko:
**Motivation**
`local_inference.py` started as notebook-support code in `notebooks/notebook_utils/`, but it has become cumbersome: it drives the workshop notebooks (Ghana, Tunisia, Brazil), the local classification demo, and local/spatial inference tooling. Living outside the package meant it was only importable via path hacks, invisible to CI, and completely untested.

**What changed**

- Moved `notebooks/notebook_utils/local_inference.py` → `src/worldcereal/local_inference.py`; imports rerouted in the 5 affected notebooks (`from worldcereal.local_inference import ...`).
- `subset_ds_temporally` rewritten around `worldcereal.train.seasonal` primitives (`align_to_composite_window` / `enumerate_composite_slots` / `advance_composite_slot`) so inference and training agree on slot semantics by construction. This also brings dekad support via a new `timestep_freq` parameter (previously months-only).
- New `pad_to_max` option + `_widen_slots_to` helper: when a unified S1+S2 season window covers fewer than `max_timesteps` slots (e.g. an 11-month union of two seasons), the selection is widened to exactly `max_timesteps` using real cube slots only — growing tail-first, falling back to the head at data boundaries, and raising if the cube can't supply enough. It never fabricates nodata timesteps.
- Removed dead `reconstruct_dataset`.

**Tests**
New `tests/worldcerealtests/test_local_inference.py` covering the temporal-subsetting contract: season-union without padding, multi-year trimming, widening toward tail/head, boundary fallback, refusal when real slots run out, dekad-grid windows, nodata-filling within coverage, and the coverage-threshold error.

**Compatibility**
Import path change is the only breaking bit: `notebook_utils.local_inference` → `worldcereal.local_inference` (all in-repo notebooks updated).


jdegerickx (approved):
Looks good to me!
At least I can confirm the training notebooks still run as expected, but they didn't use the temporal subsetting logic so I haven't tested fully.

PR #741, opened 2026-07-23
Replace per-class hard exclusion with Laplace pseudo-count smoothing

kvantricht:
**Problem**

`_get_per_bin_class_weights` and `_build_per_class_weight_grid` both had a hard threshold (`min_samples_per_class_per_bin`, default `max(1, min_per_bin // 10) = 5`): any class with fewer than 5 samples in a spatial bin was silently dropped from the per-bin weight computation and its samples received the global class weight instead.

This caused a structural problem for rare classes like `built_up` in agricultural bins. A bin might have 500 `temporary_crops` samples and 2 `built_up` samples — the local class relationship (250:1 imbalance) was invisible to the sampler. `built_up` got a global weight that didn't reflect how underrepresented it was *locally*. This is one of the contributing factors to systematic cropland commission over built-up surfaces: the model sees very few `built_up` examples per bin relative to how common they are in full-area inference.

**Solution**

Replace the hard exclusion with **Laplace (additive) pseudo-count smoothing**. The existing `min_samples_per_class_per_bin` parameter is repurposed as the pseudo-count: before calling `get_class_weights`, every class present in the bin gets `pseudo` added to its observed count:

```python
pseudo = min_samples_per_class_per_bin   # default 5

smoothed_counts = {str(lbl): int(cnt) + pseudo for lbl, cnt in class_counts.items()}
bin_w_dict = get_class_weights(..., counts_override=smoothed_counts)
```

**Effect on the problematic case** — bin with `temporary_crops`=500, `built_up`=2, `pseudo`=5:

| Class | Raw | Smoothed | `balanced` weight (pre-clip) |
|---|---|---|---|
| `temporary_crops` | 500 | 505 | ≈ 0.51 |
| `built_up` | 2 | 7 | ≈ 36.6 → clipped to `clip_max` |

`built_up` now participates at the maximum clip rate *locally*, driven by its local scarcity relative to `temporary_crops`. Previously it fell back to the global weight, which might coincidentally be similar in magnitude but carried no spatial context.

**Properties:**
- No class present in a dense bin is ever excluded from per-bin weighting.
- A class with many real samples (e.g. 500) is barely affected by a pseudo-count of 5 (≈ 1% change in effective count).
- A class with very few samples (e.g. 2) is pulled toward equal weight with the majority, and the exact degree of pull is controlled by `pseudo` — higher values make rare classes converge more strongly toward uniform weight; lower values preserve more local signal.
- The final `clip_range` is always applied, so no class can dominate regardless of how rare it is locally.
- **No API change**: `min_samples_per_class_per_bin` keeps the same name, default derivation (`max(1, min_per_bin // 10)`), and call signature. Callers that set it explicitly get the same value used as the pseudo-count.

**Files changed**
- datasets.py: `_get_per_bin_class_weights`, `_build_per_class_weight_grid`

### DIFF
--- a/src/worldcereal/train/datasets.py
+++ b/src/worldcereal/train/datasets.py
@@ -412,11 +412,11 @@ def get_class_weights(
         logger.info(f"Clipping weights to range {clip_range}")
         weights = np.clip(weights, clip_range[0], clip_range[1])
 
-    rounded = np.round(weights, 3)
-    result = {cls: float(weight) for cls, weight in zip(classes, rounded.tolist())}
+    result = {cls: float(weight) for cls, weight in zip(classes, weights.tolist())}
     if verbose:
         display = {
-            str(cls): float(weight) for cls, weight in zip(classes, rounded.tolist())
+            str(cls): round(float(weight), 3)
+            for cls, weight in zip(classes, weights.tolist())
         }
         prefix = f"[{pool_name}] " if pool_name else ""
         logger.info(f"{prefix}Class weights ({method}): {display}")
@@ -676,13 +676,20 @@ def _get_per_bin_class_weights(
                 bin_labels,
                 method=method,
                 clip_range=None,
-                normalize=True,
+                normalize=False,
                 verbose=False,
                 counts_override=smoothed_counts,
             )
         )
+        sample_mean = (
+            sum(
+                int(cnt) * bin_w_dict[str(lbl)]
+                for lbl, cnt in class_counts.items()
+            )
+            / float(class_counts.sum())
+        )
         weights[mask] = np.array(
-            [bin_w_dict[str(lbl)] for lbl in bin_labels],
+            [bin_w_dict[str(lbl)] / sample_mean for lbl in bin_labels],
             dtype=np.float64,
         )
 
@@ -882,16 +889,23 @@ def _build_per_class_weight_grid(
         smoothed_counts = {
             str(lbl): int(cnt) + pseudo for lbl, cnt in class_counts.items()
         }
-        bin_weights[(li, lo)] = _stringify_weight_dict(
+        wdict = _stringify_weight_dict(
             get_class_weights(
                 bin_labels,
                 method=method,
                 clip_range=None,
-                normalize=True,
+                normalize=False,
                 verbose=False,
                 counts_override=smoothed_counts,
             )
         )
+        sample_mean = (
+            sum(int(cnt) * wdict[str(lbl)] for lbl, cnt in class_counts.items())
+            / float(class_counts.sum())
+        )
+        bin_weights[(li, lo)] = {
+            cls: w / sample_mean for cls, w in wdict.items()
+        }
     n_total_bins = bin_groups.ngroups
     n_dense_bins = len(bin_weights)
 
--- a/src/worldcereal/train/datasets.py
+++ b/src/worldcereal/train/datasets.py
@@ -605,13 +605,14 @@ def _get_per_bin_class_weights(
     weights — within-bin counts are too sparse to estimate a stable class
     distribution otherwise.
 
-    Within a bin, classes with fewer than *min_samples_per_class_per_bin*
-    samples are excluded from the within-bin weight computation entirely
-    (they don't count toward ``num_classes`` either, removing the
-    destabilising effect of rare classes on dense ones via the ``k_bin``
-    factor in the ``balanced`` recipe). Samples of those filtered classes
-    inherit the global class weight individually. ``None`` (default)
-    auto-derives as ``max(1, min_samples_per_bin // 10)``.
+    Within a dense bin, every class present receives a Laplace (additive)
+    pseudo-count of *min_samples_per_class_per_bin* before class weights are
+    computed.  This means classes with few local samples are never excluded;
+    instead their per-bin weight is smoothly pulled toward equal weight
+    relative to majority classes, while remaining locally meaningful.  The
+    final weight is always clipped to *clip_range* so no class can dominate.
+    ``None`` (default) auto-derives the pseudo-count as
+    ``max(1, min_samples_per_bin // 10)``.
     """
     if labels.shape != bins.shape:
         raise ValueError(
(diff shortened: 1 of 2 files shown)


cbutsko:
I think this PR reveals a deeper issue than the one it addresses. The `pseudo`-count smoothing does let extremely minor classes participate in _local_ weighting - but the instability actually lives in the **per-bin normalizer**, which the smoothing doesn't touch. With the `effective` method a class's raw weight is essentially `bin_size / count`, so whichever class is smallest dominates the *unweighted class mean* we normalize by - and the PR, by un-excluding those classes, hands them the
denominator for the entire bin.

Real cell, H3-L2 `821e37…`, Czechia/Austria, N = 366,443. Raw class weights (≈ bin_size/count):
| class | n | raw weight |
|---|---:|---:|
| temporary_crops | 162,518 | 3 |
| grasslands | 118,887 | 4 |
| temporary_grasses | 39,558 | 10 |
| shrubland | 17,021 | 22 |
| permanent_crops | 16,193 | 23 |
| bare_sparsely_veg. | 7,860 | 47 |
| trees | 2,156 | 170 |
| water | 1,291 | 283 |
| built_up | 955 | 382 |
| wetlands | 4 (+5) | 40,716 |

So far so good - inverse-frequency weights are supposed to look like this. The trouble begins at the normalizer:
* **current version**: class-mean over the 9 included classes (wetlands excluded by the `min_samples_per_class_per_bin` threshold) → **105**
* **PR version**: class-mean over all 10 → **4,166** — wetlands alone contributes 98% of it
* **proposed version**: sample-weighted mean, Σ n·w / N → **10.0** - each class counts in proportion to the samples that actually carry it, so a 4-sample class can only move the normalizer by four samples' worth of mass and the denominator stays ≈ the number of classes no matter how small any class gets

Per-bin normalized weights (raw ÷ normalizer), then the global cross-bin rescale(÷ μ, which is ×34 for current, ×66 for the PR, and ×1.0 — i.e. a no-op — for the proposal) and clip (0.1, 4.0):

| class | norm old | norm PR | norm proposed | final old | final PR | final proposed |
|---|---:|---:|---:|---:|---:|---:|
| temporary_crops | 0.026 | 0.0007 | 0.280 | 0.904 | 0.100 | 0.280 |
| grasslands | 0.034 | 0.0009 | 0.362 | 1.139 | 0.100 | 0.362 |
| temporary_grasses | 0.093 | 0.002 | 0.981 | 3.115 | 0.131 | 0.981 |
| shrubland | 0.209 | 0.005 | 2.212 | 4.000 | 0.328 | 2.212 |
| permanent_crops | 0.220 | 0.006 | 2.322 | 4.000 | 0.394 | 2.322 |
| bare_sparsely_veg. | 0.448 | 0.011 | 4.729 | 4.000 | 0.722 | 4.000 |
| trees | 1.619 | 0.041 | 17.1 | 4.000 | 2.690 | 4.000 |
| water | 2.701 | 0.068 | 28.4 | 4.000 | 4.000 | 4.000 |
| built_up | 3.650 | 0.092 | 38.4 | 4.000 | 4.000 | 4.000 |
| wetlands | *excluded* | 9.77 | 4,089 | 4.000 | 4.000 | 4.000 |

(Aside: at the PR's norm stage, `get_class_weights` rounds returned weights to 3 decimals — crops' 0.0007 becomes 0.001, a +43% error, masked here only because the clip floors it anyway. Worth removing the rounding from return values regardless.)

Three takeaways:

1. **All three variants give wetlands exactly 4.0.** The rare class this PR was written for ends up identical either way - the entire difference between the three pipelines is what happens to the *other* 366,439 samples in the bin.
2. **The drag is not PR-specific - it's our current regime.** The exclusion threshold only protects against counts < 5; a *natural* 6-sample class does the same damage legally.
3. **Proposed fix**: keep the pseudo-count (as a shrinkage prior on small-count weights - it's genuinely useful in small bins), but normalize each bin by the sample-weighted mean instead of the class mean, with unrounded weights. This fixes both the PR's side effect and the pre-existing floor-pinning, is robust across `balanced`/`log`/`effective` and H3 L1/L2, and makes weights directly interpretable as "× the bin-average draw rate" (so the clip range finally means what it says).

src/worldcereal/train/datasets.py
cbutsko: Should we remove rounding from here to avoid quantization of weights in very large imbalanced bins to the 0.001 grid or zeroed?
We can still retain it for logging:
```
result = {cls: float(weight) for cls, weight in zip(classes, weights.tolist())}
if verbose:
    display = {
        str(cls): round(float(weight), 3)
        for cls, weight in zip(classes, weights.tolist())
    }
```

src/worldcereal/train/datasets.py
```diff
@@ -661,45 +662,29 @@ def _get_per_bin_class_weights(
...
-        per_sample = np.array(
-            [
-                (
-                    bin_w_dict[str(lbl)]
-                    if lbl in well_represented
-                    else global_w_dict[str(lbl)]
-                )
-                for lbl in bin_labels
-            ],
+        weights[mask] = np.array(
```
cbutsko, line 691: Suggested improvement: normalize by the **SAMPLE**-weighted mean, not the plain class mean. Each class counts in proportion to the samples that actually carry it, so a 4-sample class can only move the normalizer by 4 samples' worth of mass. Raw counts here on purpose: smoothed counts set the rates, real counts set the budget.
```
sample_mean = (
    sum(int(cnt) * bin_w_dict[str(lbl)]
        for lbl, cnt in class_counts.items())
    / float(class_counts.sum())
)
weights[mask] = np.array(
    [bin_w_dict[str(lbl)] / sample_mean for lbl in bin_labels],
    dtype=np.float64,
)
```

src/worldcereal/train/datasets.py
```diff
@@ -901,19 +879,17 @@ def _build_per_class_weight_grid(
...
-        )
-        if not well_represented:
-            continue
-        kept = bin_labels[np.isin(bin_labels, list(well_represented))]
+        smoothed_counts = {
+            str(lbl): int(cnt) + pseudo for lbl, cnt in class_counts.items()
+        }
         bin_weights[(li, lo)] = _stringify_weight_dict(
             get_class_weights(
-                kept,
```
cbutsko, line 912: Suggested change with sample-weighted mean, not the plain class mean:
```
wd = _stringify_weight_dict(
    get_class_weights(
        bin_labels, method=method, clip_range=None,
        normalize=False, verbose=False, counts_override=smoothed_counts,
    )
)
sample_mean = (
    sum(int(cnt) * wd[str(lbl)] for lbl, cnt in class_counts.items())
    / float(class_counts.sum())
)
bin_weights[(li, lo)] = {k: v / sample_mean for k, v in wd.items()}
```

cbutsko (approved):
All good!

PR #749, opened 2026-08-19
Multiclass LC10 output at inference

cbutsko:
When the loaded model carries a multiclass landcover head (>2 classes), inference now emits two additional bands alongside the existing outputs:

- `landcover_classification` — uint8 argmax over the LC classes
- `landcover_probability` — uint8-scaled winning-class probability

Models with a binary landcover head produce byte-identical output to before — the new bands are only added when the head actually has a multiclass legend, so all existing production models and workflows are unaffected.

This PR enables LC10 output for local inference and internal inspection only. It is intentionally not promoted to the UDP / Processing Hub and not part of any published product.


src/worldcereal/openeo/inference.py
```diff
@@ -1561,6 +1561,9 @@ def _format_outputs(
 
         band_layers: List[Tuple[str, np.ndarray]] = []
         cropland_mask_bool: Optional[np.ndarray] = None
+        # Multiclass landcover layers (argmax + winning probability), only
+        # produced when the landcover head has more than 2 classes.
```
kvantricht: update comment so it's clear it's only produced when more than 2 classes AND deliberately asked for?

src/worldcereal/openeo/inference.py
```diff
@@ -1582,6 +1585,18 @@ def _register_band(name: str, values: np.ndarray) -> None:
             preds_np = preds.numpy().reshape(height, width)
 
             landcover_classes = list(self.bundle.landcover_spec.class_names)
+            if len(landcover_classes) > 2 and not _emit_multiclass_landcover():
```
kvantricht: the UDF is already very long; maybe this pathway with debug message is not needed?

src/worldcereal/openeo/inference.py
```diff
@@ -2082,6 +2106,17 @@ def _normalize_udf_season_masks(value: Any) -> Optional[np.ndarray]:
     return arr
 
 
+def _emit_multiclass_landcover() -> bool:
+    """Hidden switch, deliberately NOT a job parameter: the multiclass
```
kvantricht: suggest to condense docstring to not blow this simple check up too much

kvantricht (changes requested):

src/worldcereal/openeo/inference.py
cbutsko: addressed here https://github.com/WorldCereal/worldcereal-classification/pull/749/commits/299b2b873e9b09b767423a6248ab6023db417d3c

src/worldcereal/openeo/inference.py
cbutsko: removed it https://github.com/WorldCereal/worldcereal-classification/pull/749/commits/299b2b873e9b09b767423a6248ab6023db417d3c

src/worldcereal/openeo/inference.py
cbutsko: done https://github.com/WorldCereal/worldcereal-classification/pull/749/commits/299b2b873e9b09b767423a6248ab6023db417d3c

kvantricht (approved):
Looks good!

PR #750, opened 2026-08-20
Cropcalendar fetching

jdegerickx:
This PR switches crop calendar handling from the global GeoTiffs to the global geoparquet lookup, which is easier to update.

We no longer rely purely on DOY's, which are always tricky to handle, but switched to a more clear notation using either explicit dates or dekads --> doy's, dekads and dates are now all contained within the parquet.

More explanation on dekads: we count dekads on a 3 year scale
1 - 36 = year -1
37 - 72 = year
73 - 108 = year + 1
--> year represents the target year (e.g. 2024 for global production)
This ensures that for neighbouring pixels, when running production for a given year, we have full control over which cross-calendar seasons are selected (not guaranteed for pure DOY approach).

During preparation of the global parquet, we already ensured that for each location, seasons always start on first of the month and end last day of the month.

Impacted methods in datasets.py (for training) and inference (get_season_dates_for_extent) have been updated accordingly.

All functionality to read seasonality has been moved from datasets.py to seasons.py.

IMPORTANT REMARK:
The cropcalendars for Europe have already been carefully reviewed.
The global parquet has also already been updated for other locations of the globe, but removal of obvious calendar errors has not been done yet.
Continuous revisions of calendars for different regions will still be done and the global parquet file will be updated accordingly


src/worldcereal/train/datasets.py
```diff
@@ -30,14 +29,12 @@
 )
 from torch.utils.data import Dataset, Sampler
 
-from worldcereal.data.cropcalendars import (
-    SEASONALITY_COLUMN_MAP,
-    SEASONALITY_LAT_RANGE,
-    SEASONALITY_LON_RANGE,
-    SEASONALITY_LOOKUP_COLUMNS,
-    load_seasonality_lookup,
+from worldcereal import seasons as _seasons_module
```
kvantricht: why is this imported as private module?

src/worldcereal/train/datasets.py
```diff
@@ -30,14 +29,12 @@
...
+from worldcereal.data import cropcalendars as _cropcalendars
```
kvantricht: same question.

src/worldcereal/seasons.py
```diff
@@ -1,309 +1,528 @@
...
-class NoSeasonError(Exception):
-    pass
+def _ensure_seasonality_lookup_table() -> pd.DataFrame:
+    """Load and cache the seasonality lookup table indexed by lat/lon centers."""
 
+    global _SEASONALITY_LOOKUP_TABLE
+    if _SEASONALITY_LOOKUP_TABLE is not None:
+        return _SEASONALITY_LOOKUP_TABLE
 
-class SeasonMaxDiffError(Exception):
```
kvantricht, line 21: how is it then now dealt with if a user wants to see a season for a too large spatial extent?

src/worldcereal/seasons.py
```diff
@@ -1,309 +1,528 @@
...
-from worldcereal import SEASONAL_MAPPING, SUPPORTED_SEASONS
+from worldcereal import SUPPORTED_SEASONS
 from worldcereal.data import cropcalendars
 
-# from worldcereal.utils import aez as aezloader
-from worldcereal.utils.geoloader import load_reproject
+_SEASONALITY_LOOKUP_TABLE: Optional[pd.DataFrame] = None
 
 
-class NoSeasonError(Exception):
```
kvantricht, line 17: How is now dealt with requesting seasons over ocean?

src/worldcereal/seasons.py
```diff
@@ -1,309 +1,528 @@
...
+        lat=lat,
+        lon=lon,
+        parameter="dekad",
+        fallback_to_nearest=fallback_to_nearest,
+    )
 
-    Optimized to return integer arrays for maximum efficiency. Missing/nodata
-    values are represented as 0 - filter these out with array[array > 0].
+    if sos_value > 108 or eos_value > 108:
+        logger.warning(
```
kvantricht, line 291: what is this based on? What's the rationale?

src/worldcereal/seasons.py
```diff
@@ -1,309 +1,528 @@
...
-    """
 
-    year = after_date.year
-    # DOY is 1-based so day 1 should correspond to January 1st
-    doy_date = pd.to_datetime(f"{year}-01-01") + pd.Timedelta(days=doy - 1)
+def get_season_dates_for_extent(
+    extent: BoundingBoxExtent,
+    year: int,
+    season: str = "tc-annual",
+    max_dekad_difference: int = 7,
```
kvantricht, line 425: what is this based on? Why does a user need to count in dekads to specify max variation in seasonality? A dekad also has variable length so how deterministic is this?

src/worldcereal/seasons.py
```diff
@@ -346,147 +565,55 @@ def season_doys_to_dates_refyear(sos: int, eos: int, ref_year: int):
...
 
-def get_season_dates_for_extent(
-    extent: BoundingBoxExtent,
-    year: int,
-    season: str = "tc-annual",
-    max_seasonality_difference: int = 60,
-) -> TemporalContext:
-    """Function to retrieve seasonality for a specific year based on WorldCereal
-    crop calendars for a given extent and season.
+def dekad_to_date(
```
kvantricht: didn't we have this kind of utility somewhere already? Maybe double check that it's not duplicated.

kvantricht (commented):
Added some comments/questions

src/worldcereal/train/datasets.py
jdegerickx: good question, I made the relevant methods public so they can be imported as such

src/worldcereal/seasons.py
jdegerickx: even before this PR, the seasonMaxDiffError was never thrown, we only issue  a warning so I removed this unused error

src/worldcereal/seasons.py
jdegerickx: in the query method of the parquet file there is a "fallback_to_nearest" flag which is by default set to True.
This means that the values of the nearest point will be returned (so always something will be returned if requesting seasons over the ocean).

In case the flag is set to False, a ValueError will be raised saying that no valid seasonality could be retrieved.

jdegerickx: see here --> https://github.com/WorldCereal/worldcereal-classification/blob/66c038b5280233ac4908409710335715817498f0/src/worldcereal/seasons.py#L338

src/worldcereal/seasons.py
jdegerickx: we switched from doy to dekads as primary unit for the calendars, so that's why it seemed more logical to have it this way.
In the new global crop calendars, SOS is always first of the month and EOS always last of the month, so we might as well use dekads...

src/worldcereal/seasons.py
jdegerickx: you mean the check itself? Just to verify that no invalid values are returned...

src/worldcereal/seasons.py
jdegerickx: the dekad_to_date function? I doubt it, because I introduced the concept of dekads (1 - 108) in this very PR.

src/worldcereal/seasons.py
kvantricht: we should probably add a max distance threshold for the fallback_to_nearest; if exceeded, we should still throw an exception.

src/worldcereal/seasons.py
kvantricht: was lacking the dekad concept which is now in the description of the PR. Sufficiently documented in the code too? So user knows what this 108 is based on.

src/worldcereal/seasons.py
kvantricht: as discussed, a little more specific function name could help here.

src/worldcereal/seasons.py
jdegerickx: done!

src/worldcereal/seasons.py
jdegerickx: I added documentation on top of the seasons.py file

src/worldcereal/seasons.py
jdegerickx: done!

kvantricht (approved):
Looks good to merge now! Maybe important to notify Gisat once this is in main, because - other than the UDPs - they would need to pull this main version in the processing hub to get the updates on seasonality suggestion.

PR #752, opened 2026-08-21
Cropcalendar updates

jdegerickx:
Some generic corrections of artefacts and unrealistic seasons across the globe.


kvantricht (approved):
let's go then

PR #756, opened 2026-08-27
Improved crop calendar fetching for training workflows

jdegerickx:
For inference we already moved from DOY's to dekads, now we do the same for the training workflow.

For a training point, the dekad values of SOS and EOS are fetched, then converted to real dates for the relevant year (linked to the label datetime):
- using only the year if no label datetime is provided
- ensuring the label is in the season if possible
- selecting the closest season if the label is outside the season.

As a result, we no longer need the functionality to fetch DOY cropcalendar values, so some functionality was removed.

In addition, I noticed there is a vectorized version available to fetch crop calendar values. This version was out-of-sync with the main implementation and got an update as well.


kvantricht (approved):
Let's go!

PR #759, opened 2026-09-02
Splitting with children

kvantricht:
Prevent train/validation/test leakage when polygon samples have `_child<N>` point samples.

- Group child samples with their parent sample_id before random and spatial splitting.
- Update legacy split_df to preserve parent groups for all split modes.
- Validate that all samples from a parent have the same class label.
- Add regression tests covering random, spatial, and legacy splitting.

### DIFF
--- a/src/worldcereal/train/data.py
+++ b/src/worldcereal/train/data.py
@@ -497,21 +497,21 @@ def spatial_train_val_test_split(
 
         # Assign bins to splits in a class-balanced way
         rng = np.random.RandomState(seed)
-        train_bins_list = []
-        val_bins_list = []
-        test_bins_list = []
+        train_bins_list: List[Any] = []
+        val_bins_list: List[Any] = []
+        test_bins_list: List[Any] = []
 
         for cls, cls_bins in class_bins.items():
-            cls_bins = np.array(cls_bins)
-            rng.shuffle(cls_bins)
+            class_bin_array = np.array(cls_bins)
+            rng.shuffle(class_bin_array)
 
-            n_cls_bins = len(cls_bins)
+            n_cls_bins = len(class_bin_array)
             n_cls_test = max(1, int(n_cls_bins * test_size))
             n_cls_val = max(1, int(n_cls_bins * val_size))
 
-            test_bins_list.extend(cls_bins[:n_cls_test])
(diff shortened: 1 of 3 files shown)


src/worldcereal/train/data.py
```diff
@@ -378,9 +434,28 @@ def spatial_train_val_test_split(
...
+    if stratify_label and stratify_label in group_df.columns:
+        inconsistent_labels = grouped[stratify_label].nunique(dropna=False)
+        if (inconsistent_labels > 1).any():
+            invalid_parents = inconsistent_labels[
+                inconsistent_labels > 1
+            ].index.tolist()
+            raise ValueError(
+                f"Parent samples have inconsistent '{stratify_label}' labels: "
+                f"{invalid_parents}"
+            )
```
cbutsko: a whole chunk of code seems to repeat line-by-line twice in the same file. maybe worth isolating into a common function?

src/worldcereal/utils/refdata.py
```diff
@@ -1020,21 +1020,38 @@ def split_df(
     val_size: Optional[float] = None,
     train_only_samples: Optional[List[str]] = None,
 ) -> Tuple[pd.DataFrame, pd.DataFrame]:
+    if "sample_id" not in df.columns:
+        raise ValueError("Splitting requires a 'sample_id' column.")
+
+    parent_ids = (
+        df["sample_id"].astype("string").str.replace(r"_child\d+$", "", regex=True)
+    )
```
cbutsko: maybe we can reuse `_parent_sample_ids` from `data.py`? it seems to be doing exactly that.
UPD: bad idea, we'll have circular import in that case. but maybe it makes sense to move that utility here to `ref_data.py` then? It can also be used in a few places below

cbutsko (commented):
Looks good in terms of logic! I feel that some refactoring might be needed (hence my comments about duplication), but I also think it's okay if we do it as a separate effort.

kvantricht:
> Looks good in terms of logic! I feel that some refactoring might be needed (hence my comments about duplication), but I also think it's okay if we do it as a separate effort.

good points! Are the changes pushed now what you mean?

cbutsko (approved):
Looks good!

jdegerickx (approved):
looks good to me

PR #760, opened 2026-09-03
Migrate away from artifactory

kvantricht:
This PR strips away dependencies on Artifactory, anticipating its phase-out in november and fixing current failures in tests.

### DIFF
--- a/scripts/misc/legend.py
+++ b/scripts/misc/legend.py
@@ -1,71 +1,40 @@
 """
-Example script showing how to upload, download and delete the WorldCereal crop type legend from Artifactory.
+WorldCereal legend publication and download helper.
+
+Publishing is intentionally manual: this repository no longer manages storage
+credentials or uploads legends itself. To publish a land-cover/crop-type legend:
+
+1. Place the dated CSV and PDF in ``/vitodata/worldcereal/data/legend/`` using
+   ``WorldCereal_LC_CT_legend_YYYYMMDD.{csv,pdf}``.
+2. Upload both dated files and their ``latest`` aliases with an approved S3
+   client. For the configured ``cloudferro`` rclone remote, for example:
+
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.csv \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_YYYYMMDD.csv
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.csv \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_latest.csv
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.pdf \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_YYYYMMDD.pdf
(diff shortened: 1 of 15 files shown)


scripts/misc/legend.py
```diff
@@ -1,71 +1,40 @@
...
+   client. For the configured ``cloudferro`` rclone remote, for example:
+
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.csv \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_YYYYMMDD.csv
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.csv \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_latest.csv
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.pdf \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_YYYYMMDD.pdf
+   rclone copyto WorldCereal_LC_CT_legend_YYYYMMDD.pdf \
+       cloudferro:project_dependencies/worldcereal/WorldCereal_LC_CT_legend_latest.pdf
```
jdegerickx, line 19: I tried to access the specific link for the PDF version of the legend through my browser, but it couldn't find the file.
The CSV versions of both the LC/CT and IRR legend can be downloaded through the browser

scripts/misc/legend.py
jdegerickx, line 1: I was using this functionality to update the legend quite a lot.
Would be good to have a script for internal use with explanation on how to configure this upload procedure (how to get the proper credentials etc.).
I guess as soon as we have more clarity on the final location of the legend...

src/worldcereal/extract/utils.py
```diff
@@ -36,7 +39,10 @@ def filter(self, record):
 stream_handler.addFilter(ManagerLoggerFilter())
 
 
-S2_GRID = load_s2_grid()
+@lru_cache(maxsize=1)
+def get_s2_grid() -> gpd.GeoDataFrame:
+    """Load and cache the S2 grid when an extraction workflow needs it."""
+    return load_s2_grid()
```
jdegerickx, line 45: load_s2_grid lives in openeo-gfmap, so we need to update this dependency as well?
Do we need a new version in the pyproject.toml?

jdegerickx (commented):
see comments.
Two most important ones:
- pdf version of the legend should be reachable before we communicate changes to partners
- do we require openeo-gfmap update?

src/worldcereal/extract/utils.py
kvantricht: it was put into this function so it's not called automatically when importing this worldcereal module. Do we use it operationally?

scripts/misc/legend.py
kvantricht: so that will remain blocked until S-OPS releases that solution. If it is required now, this PR cannot be merged.

jdegerickx: not blocking for the moment

src/worldcereal/extract/utils.py
jdegerickx: posed the question to vincent to update openeo-gfmap.

jdegerickx (approved):

PR #762, opened 2026-09-08
Read disabled sensor flags in inference

kvantricht:
This PR makes a change to the inference code which now reads the model `run_config.json` and more specifically its sensor disabling flags. If a sensor was disabled during training, inference will now impose sensor masking as well by setting all data of this sensor to NODATAVLUE.


uceshaf (approved):
Just one comment [Comment](https://github.com/WorldCereal/worldcereal-classification/pull/764#discussion_r3990546199) to review, otherwise it works in combination with the other two repos the way it is setup now.

PR #763, opened 2026-09-08
use openeo-gfmap==0.4.9

kvantricht:
Updated openeo-gfmap to accomodate Artifactory phase-out.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,7 +49,7 @@ dependencies = [
   "numpy>=2.4.0",
   "openeo>=0.47.0",
   "prometheo @ git+https://github.com/WorldCereal/prometheo.git@v0.1.5",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -49,7 +49,7 @@ dependencies = [
   "numpy>=2.4.0",
   "openeo>=0.47.0",
   "prometheo @ git+https://github.com/WorldCereal/prometheo.git@v0.1.5",
-  "openeo-gfmap @ git+https://github.com/Open-EO/openeo-gfmap.git@maintenance/0.4.x",
+  "openeo-gfmap @ git+https://github.com/Open-EO/openeo-gfmap.git@v0.4.9",
```
VincentVerelst: I've released `0.4.9` on PyPI, so you can just set the dependency to `openeo-gfmap==0.4.9`, without using Github as an intermediary

VincentVerelst (changes requested):

VincentVerelst (approved):

Issue #758, opened 2026-09-02
Expand `SensorMaskingConfig` and retrain global model


kvantricht:
SensorMaskingConfig currently has:

```
s1_full_dropout_prob: float = 0.0
s1_timestep_dropout_prob: float = 0.0
s2_cloud_timestep_prob: float = 0.0
s2_cloud_block_prob: float = 0.0
s2_cloud_block_min: int = 2
s2_cloud_block_max: int = 5
meteo_timestep_dropout_prob: float = 0.0
dem_dropout_prob: float = 0.0
```

Default values now are:
```
# Masking parameters
s1_full_dropout = 1.0 if args.disable_s1 else 0.2
s1_ts_dropout = 0.0 if args.disable_s1 else 0.15  # redundant when full=1.0
s2_cloud_ts = 1.0 if args.disable_s2 else 0.2
s2_cloud_block = 0.0 if args.disable_s2 else 0.15  # block not needed when ts=1.0
meteo_dropout = 1.0 if args.disable_meteo else 0.15
masking_config = SensorMaskingConfig(
    enable=enable_masking,
    s1_full_dropout_prob=s1_full_dropout,
    s1_timestep_dropout_prob=s1_ts_dropout,
    s2_cloud_timestep_prob=s2_cloud_ts,
    s2_cloud_block_prob=s2_cloud_block,
    s2_cloud_block_min=1,
    s2_cloud_block_max=5 if timestep_freq == "month" else 12,
    meteo_timestep_dropout_prob=meteo_dropout,
    dem_dropout_prob=0.1,
    seed=DEFAULT_SEED,
)
```

We need:

- [x]  to update the `SensorMaskingConfig` and its handling to also have `s2_full_dropout` (default in finetuning 0.1?) and `meteo_full_dropout` (default in finetuning 0.2?).
- [x] `disable_s2` and `disable_dem` should be added to finetuning script.
- [x] Investigate the need for a safeguard to make sure _any_ token still remains?
- [x] Retrain the global model (taking latest crop calendar and outlier changes into account).
- [x] Update `merge_head_runs.py` to build in a safeguard (shoutout to SILO, awesome series) to make sure two separately trained heads have to be compatible in terms of their disabling flags!
- [x] Update `merge_head_runs.py` to also export a sanitized version of the model ZIP without sensitive details.

uceshaf:
@kvantricht The two branches for masking are already in main.
[...]

PR #764, opened 2026-09-09
Robust masking

uceshaf:
The masking config plus the guardrails added from https://github.com/WorldCereal/worldcereal-classification/issues/758
Tested on one run of global model and the masking config works fine. 

This goes in combination with https://github.com/WorldCereal/worldcereal-finetune/tree/robust_masking

The update to merge_heads workflow will be done from another branch where we already have the sanitization scripts, and will be merged from there.


uceshaf:
@cbutsko shortened the comments in both repo branches, so once review for this repo is complete, will merge both to main

src/worldcereal/train/finetuning_utils.py
```diff
@@ -1295,7 +1296,13 @@ def prepare_training_datasets(
     classes_list : Optional[List[str]], default=None
         List of class names. If None, an empty list is used. Required for multiclass task.
     masking_config : Optional[SensorMaskingConfig], default=None
```
cbutsko: Just a thought: since we now have a separate eval_masking_config, and this one is used for train, would it make sense to mirror it in the name too, like `train_masking_config`?

cbutsko (approved):
Looks good! just a minor comment about masking config name, not critical

src/worldcereal/train/finetuning_utils.py
uceshaf: Yes, would be easier to identify. Making the change now

src/worldcereal/train/datasets.py
```diff
@@ -892,11 +892,17 @@ class SensorMaskingConfig:
...
-    restored — preferring S2, and never a sensor whose full elimination is
-    intentional (s1_full_dropout_prob or s2_cloud_timestep_prob of 1.0).
+    Each sensor has a *full* dropout knob (whole sensor gone) and, where it
+    applies, a *per-timestep* knob. A full dropout probability of 1.0 switches
+    the sensor off for the whole run (``--disable_s1`` and friends).
+
+    Two invariants are enforced (see
+    :meth:`WorldCerealDataset._rescue_fully_masked_sample`): S1 and S2 are never
+    both fully masked-or-missing, and a sample never loses every encoder token.
+    Presto builds tokens from S1, S2, meteo and DEM only, so a sample with all
```
kvantricht: remove "Presto" as we want to be encoder-agnostic.

src/worldcereal/train/datasets.py
```diff
@@ -925,31 +936,71 @@ class SensorMaskingConfig:
     enable: bool = False
     s1_full_dropout_prob: float = 0.0
     s1_timestep_dropout_prob: float = 0.0
```
kvantricht, line 942: how does this (and for other sensors) now work when this is set to 1.0? If a user sets full dropout prob to 0 but this one to 1, he silently actually disables the sensor as well? Or am I not seeing that right?

src/worldcereal/train/datasets.py
```diff
@@ -925,31 +936,71 @@ class SensorMaskingConfig:
...
             raise ValueError(
                 "s2_cloud_block_min cannot be greater than s2_cloud_block_max"
             )
         if self.s2_cloud_block_max > num_timesteps:
             raise ValueError("s2_cloud_block_max cannot exceed num_timesteps")
-        if self.s1_full_dropout_prob >= 1.0 and self.s2_cloud_timestep_prob >= 1.0:
+        if self.s1_disabled and self.s2_disabled:
             raise ValueError(
-                "s1_full_dropout_prob and s2_cloud_timestep_prob cannot both be 1.0: "
+                "S1 and S2 dropout probabilities cannot both be 1.0: "
```
kvantricht: too generic message. Which probabilities?

src/worldcereal/train/datasets.py
```diff
@@ -925,31 +936,71 @@ class SensorMaskingConfig:
...
                 "every sample would end up with S1 and S2 fully masked"
             )
+        if self.s1_disabled and self.meteo_disabled and self.dem_disabled:
```
kvantricht: don't understand this logic. You want to guard against situations where these are disable _and_ S2 happens to have no valid observations? But what if we want to train a S2-only model, something that basically many people (AI2, NASA Harvest) are usually doing?

kvantricht (commented):
Added few comments. Regarding eval_masking_config I understand this is meant to mirror full sensor dropouts? But I could not find how this is managed? Does this need to be done manually in the other repo? That leaves it up to the user to maybe not be doing this correctly? One alternative could be to impose this special mirroring here, so only very specific cases do full sensor dropouts in the eval set?

src/worldcereal/train/datasets.py
uceshaf: Removed in [1c75437](https://github.com/WorldCereal/worldcereal-classification/pull/764/commits/1c7543709ce5ff18784ced1f4d7e31b0a0ebcc4b)

src/worldcereal/train/datasets.py
uceshaf: [1c75437](https://github.com/WorldCereal/worldcereal-classification/pull/764/commits/1c7543709ce5ff18784ced1f4d7e31b0a0ebcc4b) The error now lists the offending field names and values, e.g. S1 and S2 are both eliminated (s1_full_dropout_prob=1.0, s2_cloud_timestep_prob=1.0)

src/worldcereal/train/datasets.py
uceshaf: I checked, yes this wrongly rejected S2-only or S1-only models. Didn't consider earlier that we may be turning off all other inputs including DEM/Meteo, but as you said, could be needed in some cases. So, now three-way rules are removed, only S1+S2 rule remains (we need one of them to be there, right?) [1c75437](https://github.com/WorldCereal/worldcereal-classification/pull/764/commits/1c7543709ce5ff18784ced1f4d7e31b0a0ebcc4b). At the moment at least, we don't want to train models only on DEM, or Meteo or latlon!

uceshaf:
> Added few comments. Regarding eval_masking_config I understand this is meant to mirror full sensor dropouts? But I could not find how this is managed? Does this need to be done manually in the other repo? That leaves it up to the user to maybe not be doing this correctly? One alternative could be to impose this special mirroring here, so only very specific cases do full sensor dropouts in the eval set?

It was manual in the finetune repo. It is now derived inside prepare_training_datasets: and mirrored as you suggested.  If no eval config is passed, whole-sensor elimination is mirrored from the training config automatically. Probabilty knobs are not mirrored, otherwise random masking would make the checkpoint metrics noisy. An explicitly passed eval config still would override values from config.

src/worldcereal/train/datasets.py
uceshaf: Yes, Setting `s1_timestep_dropout_prob`=1.0 while `s1_full_dropout_prob`=0.0 drops every S1 timestep, so S1 is effectively off. The same holds for s2_cloud_timestep_prob and `meteo_timestep_dropout_prob`.  The only two ways to set this are writing code against `SensorMaskingConfig `directly or the finetuning scripts (which are private), and the default is `enable=False`, so a normal user, at this time, gets no masking at all.

The only difference between the two routes is what the rescue guard may do afterwards. With the per-timestep knob at 1.0, S2 and meteo are treated as intentionally off and are never restored, so in the one case where S2 also ends up fully gone along with S1 because of the per-timestep know, the guard puts a single S1 timestep back. That is pre-existing on main and a test (`test_joint_guard_restores_s1_when_s2_disabled`) depends on it.

uceshaf: @kvantricht 
On further thinking on this, and after looking at the other branch. 
There are three options that can work with this,

1. We do not allow putting 1.0 or above as probabilities on the params that can in actual disable a sensor. Seems like a weakest gaurd. 
2. We let any value, but if it is 1.0 or above, we switch their disable flag on and raise a proper warning in logs. Would be easier to implement/maintain and then we only need the disable flags during inference. The downside is that then we would have a disconnect in invocation.log and the flags in run_config.json
3. The third option is the one we have now implemented in this branch and also in disable-sensors-inference branch ([62d5244](https://github.com/WorldCereal/worldcereal-classification/pull/762/commits/62d524492cc8146a1cc4568f0d422b1a8ae6bf86)) that we check not only the disable flags, but also the probabilities and then mirror the flags/values as they were during the runs


So, the way it is implemented now as a combination of disable flags and also probability mirroring to check if some sensor got disabled due to >=1.0 probability would work. Defaults are safe, nothing is near 1.0, so no sensor is eliminated unless someone deliberately overrides thorugh choosing the cli flag and giving it a json with 1.0 values!  Even then, the current setup handles it by itself!

But your choice if you want to go for option 2. Otherwise, I think we can merge these branches?

kvantricht: Then it looks fine to go with that third option.

PR #766, opened 2026-09-11
replace artifactory links with auxdata links

jdegerickx:
Some notebooks rely on datasets that were previously offered through artifactory.
These materials have now been moved to auxdata.terrascope.be and links have been updated.


kvantricht (approved):
Looks good. In another branch you'll also re-enable your legend upload procedure? Probably those files can also go to this location instead of S3?

PR #771, opened 2026-09-16
Classif app sensor masking

jdegerickx:
This PR ensures that users in the notebooks can make use of the sensor masking strategies and eliminate entire input modalities in both model training and inference.

New feature here is that in worldcereal_preprocessed_inputs we avoid loading inputs that are not relevant. This is only done in the notebook workflow, not in the UDP context...

### DIFF
--- a/src/worldcereal/openeo/preprocessing.py
+++ b/src/worldcereal/openeo/preprocessing.py
@@ -480,6 +480,8 @@ def worldcereal_preprocessed_inputs(
     spatial_extent: Union[GeoJSON, BoundingBoxExtent, str],
     temporal_extent: TemporalContext,
     fetch_type: Optional[FetchType] = FetchType.TILE,
+    disable_s1: bool = False,
+    disable_s2: bool = False,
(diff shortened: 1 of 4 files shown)


uceshaf:
@jdegerickx With the upcoming PRs (https://github.com/WorldCereal/worldcereal-finetune/pull/40 , https://github.com/WorldCereal/worldcereal-classification/pull/772), we changed some logic for disable flags, and once they are in main, there will be errors from this branch. 
I will add a commit to make this one also aligned with the other PRs!

src/worldcereal/openeo/preprocessing.py
kvantricht, line 1: changes to this kind of functionality need to be very thoroughly tested. Probably needs live tests for all possible combinations.

jdegerickx: I've uploaded a test script allowing each combination to be tested explicitly.
Lucile has run the script and confirms all looks good.

PR #772, opened 2026-09-16
Allow to disable latlon in inference

kvantricht:
This PR reads a `disable_latlon` flag from a model config and in case its `True` it will not feed the latlons to Presto, removing the location token from the model.


uceshaf:
Would we need to update the prometheo links post lat/lon optional change?
https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/job.py#L45

I think that's the reason tests are also failing
`generate_predictor `lacks a `disable_latlon `argument, but updated inference.py always passes one, likely breaking every openEO job

Another place where url is:
https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/openeo/feature_extractor.py#L29

and 

https://github.com/WorldCereal/worldcereal-classification/blob/e6c59d71bb3349612ee7136ac5875fb2e186fd96/src/worldcereal/openeo/inference_catboost.py#L45

uceshaf:
`infer(disable_latlon=...)` is never passed by `run_seasonal_inference` 

https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/local_inference.py#L553-L559


 or by the openEO UDF 

 https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/openeo/inference.py#L1994-L2003


so users can't use the manual switch-off yet. Should we add it to both, or drop it for now?

uceshaf:
Should we perhaps log a warning when the switch-off is used on a model trained with latlon_dropout 0? Becuase that model never seen a sample without lat/lon during, so its predictions may get worse in inference?

https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/openeo/inference.py#L1334-L1341

kvantricht:
> Should we perhaps log a warning when the switch-off is used on a model trained with latlon_dropout 0? Becuase that model never seen a sample without lat/lon during, so its predictions may get worse in inference?
> 
> https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/openeo/inference.py#L1334-L1341

given that we wont allow override in "official openEO" runs and only in local inference runs, i would leave that up to the user's responsibility. Otherwise we have to gain access here to dropout probability values once again, which is something I don't like to bring in inference code.

kvantricht:
> `infer(disable_latlon=...)` is never passed by `run_seasonal_inference`
> 
> https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/local_inference.py#L553-L559
> 
> or by the openEO UDF
> 
> https://github.com/WorldCereal/worldcereal-classification/blob/099d688d534eaae403611e7f02616ea3e64601fe/src/worldcereal/openeo/inference.py#L1994-L2003
> 
> so users can't use the manual switch-off yet. Should we add it to both, or drop it for now?

as discussed, better not in openEO inference, but can you add it to the local inference as an option?

uceshaf (approved):

